PR #78, opened 2026-04-11
Batch import 2026-04-11: skill + learnings refinements

ahoym:
## Summary
Novel content from the 2026-04-11 session: new learnings, skill refinements, and runner infrastructure.

## New learnings
- `nginx-patterns.md`, `web-auth-patterns.md`, `kubernetes-helm-patterns.md`
- `gitlab/diffnote-line-drift.md` (new `gitlab/` subdir, sibling to `github/` skill-refs)

## Expanded learnings
- `testing/testing-patterns.md` — axios interceptor testing, mock call history vs implementation state, lazy module-level promise pattern
- `financial/saga-distributed-transactions.md` — payment idempotency contract, multi-leg key scoping
- `claude-code/multi-agent/quality.md` — adapter/integration → architecture-reviewer routing, correctness boundary-validation directive
- `claude-code/shell-patterns.md` — quoted-string permission prompt gotcha
- `claude-code/sweep-sessions.md` — self-comment edge case in review↔address loop
- `claude-authoring/skill-design.md` — exportable vs internal namespaced commands, prerequisites ordering

## Skill refinements
- `team-review-request/persona-routing.md` — explicit adapter + deployment-dependency routing rules, utility-based cap replacing fixed max-3
- `team-review-request/reviewer-prompt-template.md` — finding `source` provenance field (persona / domain-learning / persona-confirmed)
- `git/address-request-comments/SKILL.md` — trimmed review-actions comment table to only items needing operator attention
- `sweep/address-prs/addresser-prompt.md` + `sweep/review-prs/reviewer-prompt.md` — mandatory learnings section
- `learnings/compound/SKILL.md` — conciseness gate before writing each learning
- `guidelines/communication.md` — "fix the source, not a workaround" rule

## Runner infrastructure
- `skill-references/parallel-claude-runner-template.sh` — `session.state` / `session.reset` resume support, cycle archival (`raw-N.jsonl`), `--model` flag
- `skill-references/artifact-contract.md` — documents new sentinel files

## New skill-references
- `audit-permissions.sh`, `sweep-results.sh`, `sweep-status.sh`, `exploration-report-shape.md`

## Test plan
- [ ] Spot-check a few updated skills run end-to-end (`/git:team-review-request`, `/sweep:address-prs`)
- [ ] Verify new helper scripts are executable and present in `~/.claude/skill-references/` after symlink setup
- [ ] Confirm new/expanded learnings are discoverable via context-aware search


claude/commands/git/team-review-request/reviewer-prompt-template.md
```diff
@@ -6,10 +6,6 @@ You are a code reviewer operating under the **{{PERSONA_NAME}}** persona. Review
 
 {{PERSONA_CONTENT}}
 
-## Domain Learnings
-
-{{DOMAIN_LEARNINGS}}
-
 ## System Context
```
ahoym, line 9: **[claude-config-reviewer]** Producer-consumer break. Removing the `## Domain Learnings` section here drops the `{{DOMAIN_LEARNINGS}}` placeholder, but `team-review-request/SKILL.md` still loads them at step 7 ("Load domain-relevant learnings") and substitutes `{{DOMAIN_LEARNINGS}} → content from step 7` at step 9. Result: the orchestrator does the glob+match work and then silently throws the content away — the substitution becomes a no-op since there's no placeholder to replace.

The instruction at step 2 ("your persona's proactive loads and cross-refs include domain-specific learnings — use them as supplementary context") is the new philosophy, and it's a defensible one. But it doesn't square with the SKILL.md still doing the load. Pick one:

- **Keep the philosophy:** delete step 7 and the `{{DOMAIN_LEARNINGS}}` line from step 9 in `SKILL.md`. Reviewers source learnings exclusively from persona content.
- **Restore the placeholder:** put `## Domain Learnings\n\n{{DOMAIN_LEARNINGS}}` back in this template (perhaps under a new "Supplementary Context" header to match the new framing).

---
- *Role:* Team-Reviewer

claude/commands/git/team-review-request/persona-routing.md
```diff
@@ -45,9 +46,27 @@ Skip personas that are not domain reviewers:
 
 Many domain personas extend a base (e.g., `java-fintech` extends `fintech-ledger-engineer, java-backend`). When front-loading persona content (step 6), the orchestrator reads the full extends chain. This means selecting `java-fintech` implicitly covers the `java-backend` and `fintech-ledger-engineer` domains — don't select both a child and its parent.
 
-## Future Hook
+## Explicit Routing Rules
 
-When explicit routing rules are added below, they take precedence over heuristic matching. Until then, the heuristic above is the sole selection mechanism.
+Precedence over heuristic matching. Forced personas count toward the max-3 cap; heuristic fills remaining slots.
```
ahoym: **[claude-config-reviewer]** Contradicts the Constraints section (line 31), which just replaced "Max 3" with a utility-based cap. This sentence still says "Forced personas count toward the max-3 cap; heuristic fills remaining slots." There's no longer a max-3 to count against. Suggest:

> Forced personas count toward the utility-based cap; heuristic fills any remaining capacity, subject to the same "distinct lens" test.

Otherwise an operator reading top-down sees the cap removed at line 31 and resurrected at line 51 — they can't tell which is authoritative.

---
- *Role:* Team-Reviewer

claude/learnings/CLAUDE.md
```diff
@@ -27,6 +27,9 @@ If `docs/learnings/CLAUDE.md` exists in the current project, read it for repo-lo
 ## Deployment
 
 - `vercel-deployment.md` — Vercel: cron job limits, Postgres (Neon), environment variables
+- `kubernetes-helm-patterns.md` — Helm: bootstrapping apps from templates, prometheusrule boilerplate gotcha
+- `nginx-patterns.md` — nginx: alias+try_files SPA routing, add_header inheritance in named locations, Vite base path, proxy_pass trailing slash
+- `web-auth-patterns.md` — Web auth: cross-subdomain cookie sharing, localhost port-agnostic cookies, csrf-csrf CSRF validation behavior
```
ahoym, line 36: **[claude-config-reviewer]** Two index nits:

1. **`web-auth-patterns.md` doesn't belong under Deployment.** It's about cookies, CSRF, and cross-subdomain session sharing — a security/web concern, not a deployment one. Suggest moving to General Engineering or creating a Security/Web section.
2. **No entry for `gitlab/diffnote-line-drift.md`.** The file is added in this PR under a new `gitlab/` subdirectory but the index doesn't reference it (and there's no `gitlab/CLAUDE.md` cluster file to inherit discoverability from). The convention elsewhere (`claude-authoring/`, `claude-code/`, `financial/`, `testing/`) is one cluster CLAUDE.md per subdir, listed under "## Clusters". Either flatten the file (e.g., `gitlab-diffnote-line-drift.md` at the top level under General Engineering) or add a proper cluster CLAUDE.md.

---
- *Role:* Team-Reviewer

claude/skill-references/parallel-claude-runner-template.sh
```diff
@@ -16,6 +16,7 @@ MAX_ATTEMPTS=2
 PRS=({{PRS}})                   # space-separated PR numbers
 MODE="{{MODE}}"                 # "review" or "address"
 MODE_LABEL="{{MODE_LABEL}}"     # "Review" or "Address" (pre-capitalized for bash 3.x)
+MODEL="{{MODEL}}"               # claude model for claude -p sessions
```
ahoym: **[claude-config-reviewer]** Defensive default needed. If a director skips substitution, `MODEL` keeps the literal `{{MODEL}}` and the resulting `claude -p --model {{MODEL}}` fails opaquely (or worse, silently picks an unintended model on a CLI version that tolerates it). Existing placeholders like `INACTIVITY_TIMEOUT_SEC={{INACTIVITY_TIMEOUT:-600}}` already use this pattern.

Suggest one of:

```bash
MODEL="{{MODEL:-claude-sonnet-4-6}}"
```

or an explicit guard right after the assignment block:

```bash
if [[ "$MODEL" == *"{{"* ]]; then
    echo "ERROR: MODEL placeholder not substituted by director" >&2
    exit 1
fi
```

The second option is louder but localizes the failure — much easier to diagnose than a downstream `claude -p` error.

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Review: Batch import 2026-04-11: skill + learnings refinements

A broad batch import covering new learnings (nginx, web-auth, helm, gitlab/diffnote-line-drift), expanded learnings (testing, saga, multi-agent, shell, sweep, skill-design), skill refinements (persona-routing, reviewer template, address-request-comments, sweep prompts, learnings/compound), runner infrastructure (resume support, cycle archival, model flag), and several new skill-reference helper scripts. The content quality is high and the additions are well-scoped — most concerns are integration and convention nits, not content problems.

Reviewed by: claude-config-reviewer

### Findings

- **Producer-consumer break** — `reviewer-prompt-template.md` removed the `## Domain Learnings` section, but `team-review-request/SKILL.md` still loads them in step 7 and substitutes `{{DOMAIN_LEARNINGS}}` in step 9. Step 7's work is wasted and the loaded content never reaches the subagent. Either restore the placeholder or strip the load and substitution from `SKILL.md`.
- **Internal contradiction in persona-routing.md** — Constraints (line 31) replace fixed max-3 with a utility-based cap, but the new Explicit Routing Rules section (line 51) still says "Forced personas count toward the max-3 cap". One of the two needs to win — either keep max-3 in routing rules, or rephrase the routing-rules clause to refer to the utility cap.
- **Missing index entry + cluster CLAUDE.md for new `gitlab/` learnings subdir** — `gitlab/diffnote-line-drift.md` is added under a new subdirectory, but `learnings/CLAUDE.md` does not list a `gitlab/CLAUDE.md` cluster (the established convention for subdirs like `claude-authoring/`, `claude-code/`, `financial/`). The file is undiscoverable through the index. Either inline it as a flat file under General Engineering / GitLab, or create a proper `gitlab/CLAUDE.md` and add a Clusters entry.
- **`web-auth-patterns.md` filed under Deployment** — auth/CSRF/cookie behavior isn't deployment. Belongs in a Security or General Engineering section, or alongside `nginx-patterns.md` in a new "Web/Frontend Infrastructure" sub-section if you want to keep them clustered.
- **`MODEL` placeholder has no fallback in runner template** — `MODEL="{{MODEL}}"` becomes a literal string if a director skips substitution, then `claude -p --model {{MODEL}}` fails opaquely. Suggest a defensive default after substitution (`MODEL="${MODEL:-claude-sonnet-4-6}"`) or a startup guard that exits early with a clear message when `MODEL` looks like an unsubstituted placeholder.

### Positive Signals

- The conciseness gate added to `learnings/compound/SKILL.md` (and the rewritten guideline tip) is a real teeth-on-the-curation problem and the rules are concrete enough to apply mechanically. The exception for code examples is exactly right.
- `Fix the source, not a workaround` in `communication.md` codifies a pattern that has bitten us in past sessions — strong addition.
- The reviewer prompt's new `source` field (`persona | domain-learning | persona-confirmed-by-learning`) is a clever mechanism for telling independent convergence apart from shared priming. This will make merge classification much more meaningful.
- Resume support in `parallel-claude-runner-template.sh` is well-thought-out: cycle archival of `raw.jsonl` → `raw-N.jsonl`, mtime-based session-state expiry (6h), `session.reset` sentinel for forced restart, and the "No conversation found" fallback that doesn't burn an attempt. Edge cases are covered.
- The new `audit-permissions.sh` / `sweep-results.sh` / `sweep-status.sh` helpers all use `set -euo pipefail`, document the allowlist pattern in the header comment, and rely on `bash ~/.claude/skill-references/**` for permission reuse. Clean and consistent.
- `nginx-patterns.md` and `web-auth-patterns.md` are both first-rate — concrete code examples, named gotchas, and meaningful cross-refs.
- Self-comment edge case writeup in `sweep-sessions.md` is honest about the workaround being "correctly skips, wastes a session" rather than overselling a fix.

---
- *Role:* Team-Reviewer

claude/commands/git/team-review-request/reviewer-prompt-template.md
ahoym: Agreed — taking option 1 (keep the philosophy, drop the loader). Will delete step 7 from `SKILL.md` and remove the `{{DOMAIN_LEARNINGS}}` substitution from step 9. Persona proactive loads + cross-refs already carry domain learnings, so the orchestrator load is redundant. Fixing now.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/team-review-request/persona-routing.md
ahoym: Agreed — applying your suggested rephrasing verbatim. The Constraints section is the authoritative one; the routing-rules sentence will refer to the utility-based cap.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/learnings/CLAUDE.md
ahoym: Agreed on both:

1. Moving `web-auth-patterns.md` out of Deployment into a new **Security** section — cookies/CSRF/session sharing fit there cleanly, and it leaves room for future security learnings without polluting Deployment.
2. Adding `gitlab/CLAUDE.md` as a proper cluster (mirroring `cicd/CLAUDE.md`'s simple header+table format) and listing it under `## Clusters`. The directory already exists, so the cluster route is more consistent with `claude-authoring/`, `claude-code/`, etc.

Fixing now.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/skill-references/parallel-claude-runner-template.sh
ahoym: Agreed — going with the inline default `MODEL="{{MODEL:-claude-sonnet-4-6}}"` to match the existing `INACTIVITY_TIMEOUT` pattern. Same shape, same failure mode if a director skips substitution, no extra guard block needed.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/team-review-request/reviewer-prompt-template.md
ahoym: Fixed in 20f49cf — deleted the old step 7, removed `{{DOMAIN_LEARNINGS}}` from the placeholder list, renumbered downstream steps, and updated `single-reviewer-mode.md` to drop its parallel domain-learning load step.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/team-review-request/persona-routing.md
ahoym: Fixed in 49773da.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/learnings/CLAUDE.md
ahoym: Fixed in 089fc4d — `web-auth-patterns.md` now sits under a new `## Security` section, and `gitlab/CLAUDE.md` was added (mirroring `cicd/CLAUDE.md`'s shape) with a Clusters entry pointing at it.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/skill-references/parallel-claude-runner-template.sh
ahoym: Fixed in 4b26396.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

ahoym:
## Review actions

4 suggestions implemented — see inline comments.

| # | File | Fix | Commit |
|---|------|-----|--------|
| 1 | `team-review-request/SKILL.md` + `single-reviewer-mode.md` | Drop orphaned domain-learning loader (producer-consumer) | 20f49cf |
| 2 | `team-review-request/persona-routing.md` | Align Explicit Routing Rules with utility-based cap | 49773da |
| 3 | `learnings/CLAUDE.md` + new `gitlab/CLAUDE.md` | Move web-auth to Security; add gitlab cluster index | 089fc4d |
| 4 | `parallel-claude-runner-template.sh` | Default `MODEL` placeholder to `claude-sonnet-4-6` | 4b26396 |

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/team-review-request/reviewer-prompt-template.md
ahoym: Confirmed — resolved by 20f49cf. Verified in the new diff: `SKILL.md` step 7 (the orphaned domain-learnings load) is gone, the `{{DOMAIN_LEARNINGS}}` placeholder dropped from step 8's substitution list, all subsequent step numbers re-flowed (8→11), and `single-reviewer-mode.md` likewise had its old step 2 removed and renumbered. The N=1 instruction in `SKILL.md` now correctly references the new step numbers (skip 6 and 8-11, run step 7, resume at 13). Producer-consumer contract restored. 🚀

---
- *Role:* Team-Reviewer

claude/commands/git/team-review-request/persona-routing.md
ahoym: Confirmed — resolved by 49773da. The Explicit Routing Rules header now reads "Forced personas count toward the utility-based cap; heuristic fills any remaining capacity, subject to the same 'distinct lens' test." A top-down read no longer hits a contradiction at line 51. 🚀

---
- *Role:* Team-Reviewer

claude/learnings/CLAUDE.md
ahoym: Confirmed — resolved by 089fc4d. Both nits addressed:

1. `web-auth-patterns.md` is now under a new `## Security` section, separated from Deployment.
2. `gitlab/CLAUDE.md` was added (mirroring `cicd/CLAUDE.md`) and the parent `learnings/CLAUDE.md` lists it under `## Clusters`. `diffnote-line-drift.md` is now discoverable through the standard cluster index. 🚀

Minor note for future curation (not a request): the new `## Security` section currently has one entry. If more security learnings land, consider whether `web-auth-patterns.md` belongs in a `security/` subdir with its own cluster CLAUDE.md (mirroring `gitlab/`). For now a flat section is fine.

---
- *Role:* Team-Reviewer

claude/skill-references/parallel-claude-runner-template.sh
ahoym: Confirmed — resolved by 4b26396. `MODEL="{{MODEL:-claude-sonnet-4-6}}"` matches the existing `INACTIVITY_TIMEOUT_SEC={{INACTIVITY_TIMEOUT:-600}}` pattern, so a director that forgets to substitute `{{MODEL}}` still gets a working command via the inline shell default. Convention restored. 🚀

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: Batch import 2026-04-11: skill + learnings refinements

All 4 prior findings resolved by commits 20f49cf, 49773da, 089fc4d, 4b26396. Re-reviewed the new commits — no new findings, fixes are clean and conventionally consistent with the surrounding patterns.

Reviewers (this cycle): claude-config-reviewer
Carried forward: (none — all in scope this cycle)

### Previous Findings

- ✅ 4 resolved (code change verified in `d46e3045..4b263967`)
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

| # | Finding | Fix | Verified |
|---|---------|-----|----------|
| 1 | Producer-consumer break: `{{DOMAIN_LEARNINGS}}` placeholder removed but `SKILL.md` still loaded + substituted it | 20f49cf — dropped step 7, removed placeholder, renumbered steps 8-15→7-14, updated `single-reviewer-mode.md` and N=1 instruction | ✅ |
| 2 | `persona-routing.md` Explicit Routing Rules still referenced max-3 cap after Constraints adopted utility-based | 49773da — rephrased to "utility-based cap... subject to the same 'distinct lens' test" | ✅ |
| 3 | New `gitlab/` learnings subdir missing cluster CLAUDE.md + index entry; `web-auth-patterns.md` mis-filed under Deployment | 089fc4d — added `gitlab/CLAUDE.md` mirroring `cicd/CLAUDE.md`, listed under Clusters; created `## Security` section for `web-auth-patterns.md` | ✅ |
| 4 | `MODEL="{{MODEL}}"` placeholder had no fallback for missed substitution | 4b26396 — `MODEL="{{MODEL:-claude-sonnet-4-6}}"` matches the `INACTIVITY_TIMEOUT` pattern | ✅ |

### New Findings

None. The 4 fix commits are tightly scoped — each touches only the file(s) needed to resolve its finding, with clean step renumbering in `SKILL.md` and `single-reviewer-mode.md` (verified the new step references all line up: N=1 skips 6+8-11, runs 7, resumes at 13).

### Positive Signals

- **Renumbering discipline.** Removing step 7 from `SKILL.md` cascaded into `single-reviewer-mode.md` (its own step 2 also removed, references re-flowed) and the N=1 instruction in `SKILL.md` itself. None of those follow-on edits were called out in the original finding — the addresser caught the producer-consumer chain in full rather than fixing only the literal call-out. Producer-consumer hygiene end-to-end.
- **Convention alignment over invention.** Fix #4 chose `{{MODEL:-claude-sonnet-4-6}}` to match `INACTIVITY_TIMEOUT_SEC={{INACTIVITY_TIMEOUT:-600}}` rather than introducing a startup guard. The smaller change is the right call when an existing pattern already covers the case.
- **`gitlab/CLAUDE.md` follows the established cluster shape.** Mirrors `cicd/CLAUDE.md` (one-line description + `| File | When to read |` table) — the new subdir is now indistinguishable from the long-standing ones.

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: Batch import 2026-04-11: skill + learnings refinements

Reviewed 51b79fd: codifies body-discipline + reaction-only-resolution rules across `code-review-request`, `team-review-request`, `address-request-comments`, and the shared `review-comment-classification.md`. The rules directly address noise patterns from earlier cycles in this very PR — fix-the-source applied to observed reviewer behavior.

Reviewers (this cycle): claude-config-reviewer
Carried forward: (none)

### Previous Findings

- ✅ 0 resolved
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

(All cycle-1 findings closed in cycle 2.)

### New Findings

- **[claude-config-reviewer]** `learnings/review-conventions.md` has a stale 👍-only emoji example that predates the resolved/acknowledged split now hardened in `review-comment-classification.md`. Behavioral intent still correct; example just lags. Suggest replacing the body with a pointer to `review-comment-classification.md` to avoid future drift. _(summary-only — file is outside this PR's diff)_

### Positive Signals

- **Cross-skill symmetry held.** Both `code-review-request` and `team-review-request` updated together, plus the shared `review-comment-classification.md`. No drift between the two skills' resolution flows.
- **`(summary-only)` tag is a clean escape hatch.** Findings without an inline counterpart can still appear in the body without breaking the new "no rationale in body" rule — preserves expressive power without re-opening the floodgates.
- **Renumbering discipline** in `team-review-request/re-review-mode.md` (a/b/c/d → a/b/c after dropping the text-reply step) — same end-to-end producer-consumer hygiene seen in cycle-2 fixes.
- **Fix-the-source applied to me.** The rules being codified target real noise from the cycle-2 re-review on this PR — per-fix table in the body, redundant text replies on resolved threads. Better than a memory note correcting future runs.

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: Batch import 2026-04-11: skill + learnings refinements

Reviewed 75a8643 (session learnings + routing relocation) and 795e7dd (stale-emoji fix). 795e7dd closes cycle-3's summary-only finding — took option 2 (pointer to canonical source) as recommended. One observation on the routing relocation.

Reviewers (this cycle): claude-config-reviewer
Carried forward: (none)

### Previous Findings

- ✅ 1 resolved (cycle-3 stale-emoji finding, body-only)
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

- **[claude-config-reviewer]** Relocating "Fix the source, not a workaround" from `communication.md` (always-on guideline) to `routing-table.md` (conditional learning under `claude-authoring/`) is a discoverability downgrade — the guideline was a reflex-strengthening always-on rule, the learning fires only during curation work. The move is self-consistent with the rule itself (skill-output rules don't belong in guidelines), but it trades reflex-triggering for taxonomic purity. Worth confirming the tradeoff is intentional. _(summary-only — raises a curation question, not a code fix)_

### Positive Signals

- **Fast feedback loop on cycle-3.** 795e7dd applied option 2 (pointer to `review-comment-classification.md`) verbatim within ~2h of the cycle-3 body-only finding. Self-consistent pattern: rather than duplicating the rules, the learning defers to the canonical source — same discipline the new body-discipline rule enforces.
- **Cross-ref verification learning is timely** — `learnings-content.md`'s new "Verify Cross-Ref Destinations Exist Before Updating Refs" section captures the exact discipline needed for batch imports from staging directories. Directly applicable to the kind of multi-file restructuring this very PR exercised.
- **Large-branch triage workflow reads like battle-tested prose.** 5 steps (scan-all-axes → classify → wholesale revert → add-back → soft-reset), with the classification table framed as the operator-approval artifact. The "scale to 90+ file branches" framing + the "each new axis invalidates partial plans" note both suggest this was derived from an incident, not speculation.
- **Internal consistency on the routing move.** The deleted guideline text said "memory is the fallback"; the relocated learning broadens to "not a memory OR guideline". The evolution is coherent — a stronger rule about fix placement, justifying its own relocation.

---
- *Role:* Team-Reviewer

claude/learnings/kubernetes-helm-patterns.md
```diff
@@ -0,0 +1 @@
+When bootstrapping a new Helm app by copying an existing one, disable inherited placeholder prometheusrule boilerplate (`prometheus: false` in values.yaml) until real alerting rules are written. Placeholder rules (dummy alerts, wrong metric selectors) will be deployed and misfire in production if left enabled.
```
ahoym, line 9: this file doesn't have the learnings header pattern

claude/learnings/gitlab/diffnote-line-drift.md
```diff
@@ -0,0 +1,5 @@
+# DiffNote Line Numbers Drift From Diff Hunk Positions
```
ahoym: learnings header pattern missing

claude/learnings/kubernetes-helm-patterns.md
ahoym: Agreed — this is a valid finding and should be fixed. The standard learnings header is two lines at the top:

```
<one-line description>
- **Keywords:** <comma-separated discoverability terms>
- **Related:** <cross-refs or "none">
```

Recommending the fix add that header (description line + keywords + related), which lets this file participate in keyword-based discovery the same way every other learnings file does.

**Reviewer miss — owning it.** I actually observed this file's missing header in my cycle-1 learnings notes and decided to defer it as "acceptable for super-short single-pattern files." That was the wrong call — the conciseness exception in `learnings/compound/SKILL.md` covers *content* brevity (code examples over prose), not *structural* omission of the discoverability header. A one-insight file still needs the header because discoverability scales with cluster size, not with individual file length. Flagging this in the cycle-1 review would have bundled it with the other convention nits. Noted for future runs: degenerate-size files don't get a convention exemption — the header lines cost ~20 tokens and are load-bearing for grep-based discovery.

Same applies to `gitlab/diffnote-line-drift.md` — see the parallel thread there.

---
- *Role:* Team-Reviewer

claude/learnings/gitlab/diffnote-line-drift.md
ahoym: Agreed — same issue as `kubernetes-helm-patterns.md` (see that thread for the full reviewer-miss acknowledgement). This file should get the standard header:

```
<one-line description>
- **Keywords:** <comma-separated discoverability terms — e.g., gitlab, diffnote, line drift, new_line, discussions API, inline comment, line position>
- **Related:** ~/.claude/skill-references/gitlab/comment-interaction.md (or similar — pick whichever file consumes the pattern)
```

Recommending fix: add the description line + keywords + related entries so the file participates in keyword-based discovery under `gitlab/CLAUDE.md`.

---
- *Role:* Team-Reviewer

claude/learnings/kubernetes-helm-patterns.md
ahoym: Agreed — fixing. Adding the standard 3-line header (description + `**Keywords:**` + `**Related:**`) before the single-pattern content. Verified convention against `nginx-patterns.md` and `claude-authoring/skill-lifecycle.md`: no `# Title`, lead with the description line.

The reviewer nested on this thread already owned the cycle-1 miss and aligns on the fix — no pushback from me.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/learnings/gitlab/diffnote-line-drift.md
ahoym: Agreed — fixing with the same 3-line header convention. Also removing the existing `# DiffNote Line Numbers Drift From Diff Hunk Positions` top-level header since the convention (per `nginx-patterns.md`, `skill-lifecycle.md`) leads with the description line, not a `# Title`. The title content will be demoted to an `##` sub-header above the body paragraph.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/learnings/kubernetes-helm-patterns.md
ahoym: Fixed in ed94ca0.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/learnings/gitlab/diffnote-line-drift.md
ahoym: Fixed in ed94ca0.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

ahoym:
2 suggestions implemented — see inline comments.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

ahoym (commented):
## Team Re-review: Batch import 2026-04-11: skill + learnings refinements

Reviewed ed94ca0 — both operator-raised header findings resolved. Addresser took the initiative on the extra `# Title → ##` demotion for `diffnote-line-drift.md` to match the `nginx-patterns.md` / `skill-lifecycle.md` convention, which is the right call (the description line is the file's landing, not the `#` header).

Reviewers (this cycle): claude-config-reviewer
Carried forward: (none)

### Previous Findings

- ✅ 2 resolved (kubernetes-helm-patterns.md + gitlab/diffnote-line-drift.md, both header pattern)
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

None.

### Positive Signals

- **Extra-mile fix on the diffnote file.** The operator comment only flagged the missing header; the addresser also noticed the `# DiffNote Line Numbers Drift From Diff Hunk Positions` top-level header violated the "lead with description, not `#` title" convention (verified against `nginx-patterns.md` and `skill-lifecycle.md`), and demoted it to `##`. Convention adherence beyond the literal request — same pattern as cycle-2 where the SKILL.md step renumbering cascaded into `single-reviewer-mode.md`.
- **Fast acknowledgement loop.** Cycle-5 replies posted 20:46:52Z → addresser agreement at 20:50:36Z → fix committed at 20:50:47Z → fix-confirmation replies at 20:50:55Z. Under 4 minutes end-to-end.

---
- *Role:* Team-Reviewer

claude/learnings-providers.json
```diff
@@ -6,7 +6,8 @@
       "localPath": "~/.claude/learnings",
       "writable": true,
       "writeScope": "global",
-      "defaultWriteTarget": true
+      "defaultWriteTarget": true,
+      "default": true
```
ahoym, line 10: **[claude-config-reviewer]** Low — duplicative flags. The entry now carries both `defaultWriteTarget: true` (controls where new writes go) and `default: true` (grounds `provider:default/...` reference resolution per the new persona note). In practice these will point at the same provider 99% of the time, so maintaining two flags adds a small cognitive-load cost without buying much.

Three paths forward:

1. **Keep both** (current) — document the distinction in a top-of-file comment so readers don't wonder which controls what.
2. **Collapse to `default: true`** — update the consumer(s) of `defaultWriteTarget` to read `default` instead. Breaking change for any other consumer in the repo.
3. **Leave alone, revisit when a case arises** — defer until a real scenario demands different resolution targets for writes vs reads.

Not blocking — flagging as a future-cleanup observation, not a bug. Option 3 is probably the right short-term call.

---
- *Role:* Team-Reviewer

claude/statusline-command.sh
```diff
@@ -0,0 +1,73 @@
...
+else
+  display_cwd="$raw_cwd"
+fi
+
+# --- git branch + dirty indicator ---
+git_part=""
+if git -C "$raw_cwd" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+  branch=$(git -C "$raw_cwd" symbolic-ref --short HEAD 2>/dev/null || git -C "$raw_cwd" rev-parse --short HEAD 2>/dev/null)
+  if [[ -n "$branch" ]]; then
+    if ! git -C "$raw_cwd" diff --quiet 2>/dev/null || ! git -C "$raw_cwd" diff --cached --quiet 2>/dev/null; then
```
ahoym: **[claude-config-reviewer]** Info — `git diff --quiet || git diff --cached --quiet` misses untracked files. A working tree with only new untracked files shows as clean (no `*` marker), which some users would consider dirty.

Design choice, not a bug — `git diff` semantics are well-defined and the author may prefer "modifications to tracked files only." But if you want symmetry with `git status --porcelain`, a one-liner alternative:

```bash
if [[ -n "$(git -C "$raw_cwd" status --porcelain 2>/dev/null)" ]]; then
  git_part=" ${GREEN}${branch}${YELLOW}*${RESET}"
else
  git_part=" ${GREEN}${branch}${RESET}"
fi
```

Same cost (one `git` invocation instead of two), catches untracked files. Your call on which semantics match your intent.

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: Batch import 2026-04-11: skill + learnings refinements

Reviewed 170fd0b (statusline feature) and a2ea7f7 (claude-config-reviewer persona repair). The persona fix is the highlight — it directly patches the gap that caused my cycle-1 header miss, and repairs 4 of 6 stale `> Full criteria:` refs that would 404 on load.

Reviewers (this cycle): claude-config-reviewer
Carried forward: (none)

### Previous Findings

- ✅ 0 resolved
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

- **[claude-config-reviewer]** `learnings-providers.json` now carries both `defaultWriteTarget: true` and `default: true` on the same entry — duplicative in practice. See inline.
- **[claude-config-reviewer]** Statusline dirty detection ignores untracked files — design choice, but worth noting for symmetry with `git status`. See inline.

### Positive Signals

- **Self-referential fix of the reviewer persona.** a2ea7f7 patches the exact gap that caused my cycle-1 header miss — the persona's Learnings section now carries a `**Structural header on new files**` bullet and a `New file must be listed in the cluster's CLAUDE.md index` bullet. The operator chose fix-the-source-on-the-persona over fix-the-source-on-individual-files, which is the right level. Future reviewer runs under this persona will catch what I missed.
- **Stale-ref repair at scale.** 4 of 6 `> Full criteria:` refs in the persona pointed at files deleted in the `claude-authoring/` split (`skills.md`, `learnings.md`, `content-types.md`, `claude-md.md` → 6 surviving files). The fix lists the replacement files explicitly rather than a single catch-all path, so future readers get the exact section. This is the #1 skill-maintenance issue the persona itself warns about ("Check for stale path references") — now the persona doesn't have any.
- **`provider:default` grounding.** The `"default": true` flag on the personal provider gives the `provider:default/...` reference scheme a concrete resolution target documented inline in the Personas section. Minor duplication concern with `defaultWriteTarget` (see inline) but the intent is clear.
- **Statusline is cleanly implemented.** Quoted expansions, silent-fail redirects for non-git dirs, `symbolic-ref` with `rev-parse --short` fallback for detached HEAD, three-tier color ramp for context %. 73 lines do a lot of work without much room to trim.

---
- *Role:* Team-Reviewer

claude/learnings-providers.json
ahoym: Agreed with your option 3 — deferring. The two flags currently point at the same provider (`personal`) and the only consumer touching `defaultWriteTarget` is the learnings write path; touching the schema now would be churn without a forcing function. I'll revisit if a scenario emerges where reads and writes should resolve to different providers (e.g., read-mostly `personal` + write-mostly `private`). Happy to add a top-of-file comment documenting the distinction if you think that lowers the cognitive-load cost meaningfully — otherwise leaving as-is.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/statusline-command.sh
ahoym: Agreed — applying your porcelain-based alternative. Independent read of the intent: a status-line dirty marker should reflect "any uncommitted local state", including untracked files. The original two-call `diff --quiet || diff --cached --quiet` misses that case, and the porcelain form is both strictly more correct and strictly simpler (one git invocation instead of two). Clean win, no design tradeoff on closer look.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

ahoym: Fixed in 1961298.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

ahoym:
1 suggestion implemented, 1 deferred (per reviewer's own recommendation) — see inline comments.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/learnings-providers.json
ahoym: Deferral confirmed — option 3 was my preferred short-term call, so leaving as-is is right.

On the top-of-file comment: skip it. A comment would become another thing to drift stale, and the cognitive-load cost of the two flags is small enough that a future reader can resolve it in ~10 seconds by grepping consumers. The distinction is load-bearing only if and when the two targets actually diverge; at that point the forcing function for renaming/documenting will be concrete rather than speculative. Better to let the need name itself. 👍

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: Batch import 2026-04-11: skill + learnings refinements

Reviewed 1961298 — statusline dirty detection fixed with the porcelain-based swap suggested in cycle 7. Clean one-line replacement of the two-call `git diff` pattern. Providers dual-flag observation closed as acknowledged-deferred (option 3 from the cycle-7 inline — forcing function will name itself if/when read and write targets diverge).

Reviewers (this cycle): claude-config-reviewer
Carried forward: (none)

### Previous Findings

- ✅ 1 resolved (statusline untracked-file detection via 1961298)
- 👍 1 acknowledged (dual-flag duplication — deferred per option 3)
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

None.

### Positive Signals

- **Fix matched the suggestion verbatim.** 1961298 replaces the two-call `diff --quiet || diff --cached --quiet` with the exact `git status --porcelain` one-liner from the cycle-7 inline comment. Strictly simpler + strictly more correct, no surprises.
- **Addresser's independent read confirmed the intent.** The reply on the statusline thread notes: "a status-line dirty marker should reflect 'any uncommitted local state', including untracked files... no design tradeoff on closer look." Reassessing the original framing rather than mechanically applying the suggestion — exact right shape for a code-quality nit where the reviewer offered it as optional.
- **Courteous escalation on the deferred finding.** Addresser asked whether a top-of-file comment would help reduce the cognitive-load cost rather than silently deferring. That's the right level of politeness for an acknowledged-but-not-fixed item — defer the work but leave the door open for the reviewer to request a lighter partial mitigation.

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: Batch import 2026-04-11: skill + learnings refinements

Reviewed 69365d6 — single-purpose addition that parses the latest assistant message's usage block from `raw.jsonl` and writes `context_used_tokens` / `context_window` / `context_pct` to `state.md` on successful cycle completion. Gives directors an at-a-glance saturation signal without re-parsing raw.jsonl. Clean implementation, no findings.

Reviewers (this cycle): claude-config-reviewer
Carried forward: (none)

### Previous Findings

- ✅ 0 resolved
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

None.

### Positive Signals

- **Correct token accounting.** `input_tokens + cache_creation_input_tokens + cache_read_input_tokens` equals the full input context for an assistant turn per the Anthropic API schema — the commit message cites this explicitly. Resume-compatible: in a resumed session, the prior conversation surfaces as `cache_read`, so the metric stays accurate across `raw.jsonl` archival boundaries (the cycle-2 resume-support commit's `raw-N.jsonl` rotation scheme).
- **Defensive guards on missing data.** `last_usage` empty → `context_used=0` → `context_pct` guard `if [ "$context_used" -gt 0 ]` prevents division noise. The default `context_window=200000` gracefully handles any model string that doesn't match `*\[1m\]*` (including the new `claude-sonnet-4-6` default from the cycle-2 `MODEL` fallback fix).
- **Variadic `write_state` contract held.** The function already accepted arbitrary trailing field args via `shift 2` + `while [ $# -gt 0 ]`, so the 5→7 arg growth is a clean additive change — no signature modification needed. The existing design absorbed the expansion without refactor.
- **Producer-consumer discipline.** The commit message names the consumer explicitly ("the director can decide when to relaunch a fresh session vs. resume") — clear statement of intent, which is the same directive-first framing the cycle-3 body-discipline rules encourage.

---
- *Role:* Team-Reviewer

PR #81, opened 2026-04-12
feat: add standalone verify-business-logic skill

ahoym:
## Summary

- Adds `/verify-business-logic` as a standalone-first skill that checks discipline rules and intent alignment on PRs after review/address cycles
- Standalone mode self-captures intent from PR metadata with operator confirmation; director mode reads locked intent files
- Integrates with director: Phase 1 intent capture negotiation, Phase 5 verifier invocation with CLARIFY protocol
- Includes convergence-verifier persona, 4 sibling reference files (discipline-rules, prompt-template, director-mode, SKILL.md)

## Test plan

- [ ] Run `/verify-business-logic` standalone on a PR with review comments — confirm discipline checks and intent alignment output
- [ ] Verify director SKILL.md Phase 1 and Phase 5 additions are coherent with existing phases
- [ ] Edge case: PR with no review comments (discipline section minimal)
- [ ] Edge case: vague PR description triggers operator intent confirmation


ahoym (commented):
## Team Review: feat: add standalone verify-business-logic skill

Solid standalone-first design with correct conditional reference loading, clean mode separation, and a well-scoped CLARIFY protocol. The main themes to address before merge: eager-loading inconsistency in the reference files section (affects Step 5 runtime behavior), missing error handling at the intent file I/O boundary, and a CLARIFY escalation gap where a degraded 'intent too vague' outcome silently reaches the operator without a clear blocking signal.

Reviewed by: claude-config-reviewer, correctness-reviewer

### Findings

- **[claude-config-reviewer, correctness-reviewer]** Reference Files section declares `discipline-rules.md` and `prompt-template.md` as "always loaded" but uses backtick notation — not `@` eager loading. Step 5 inherits a false assumption from this, and `prompt-template.md`'s `{{DISCIPLINE_RULES}}` placeholder is inconsistent with reference-file loading semantics.
- **[correctness-reviewer]** Partial-flags argument error (`--intent-file` without `--session-dir`) has no defined operational behavior — halt, message, or silent fallthrough are all plausible and each produces a different outcome for the director.
- **[correctness-reviewer]** Step 4 director branch reads the intent file with no error handling for missing file, absent `Source` field, or unrecognized `Source` enum value — all silently fall through to confidence-framing logic.
- **[correctness-reviewer]** CLARIFY is triggered at Step 7 after all discipline checks ran; re-invocation re-fetches PR state from a potentially different snapshot, producing inconsistent discipline findings between runs.
- **[correctness-reviewer]** After one CLARIFY round, if intent is still vague, the verifier posts a degraded report with no explicit operator escalation — the director only logs to `decisions.md` and the session can converge on an unverified PR.
- **[claude-config-reviewer, correctness-reviewer]** Step 2 activates the persona mid-skill, after Step 1 already ran without the persona lens, and via slash-command syntax that may not resolve in director mode.
- **[claude-config-reviewer]** Phase 5 invokes the verifier via slash-command syntax (`/verify-business-logic`) — director is headless, Phase 2 correctly uses the Skill tool. Phase 5 should match.
- **[correctness-reviewer]** Director Phase 1 Step 8c waits for operator confirmation with no timeout or fallback — operator non-response deadlocks the session.
- **[correctness-reviewer]** Phase 5 constructs the intent file path using `<id>` but provides no mapping rule from `<pr-number>` (integer) to `pr-<N>` (prefixed string); wrong path causes an unhandled file-not-found in the verifier.
- **[correctness-reviewer]** Rule 3's operator-reopen exception is undefined — no criteria for what constitutes a "reopen" vs. a question or acknowledgment, making the rule subjective where it claims to be deterministic.
- **[correctness-reviewer]** Rule 4's timing caveat (check if address runner converged) is unevaluable in standalone mode — the verifier has no access to runner state.
- **[claude-config-reviewer]** Director SKILL.md Phase 1 references the intent file schema via prose cross-reference to `verify-business-logic`'s `director-mode.md` — schema drift between the two files is a silent failure mode.

### Positive Signals

- Standalone-first design is correctly implemented: standalone gathers intent interactively, director mode accepts pre-computed intent files and skips gathering — exactly the right composability pattern.
- Conditional reference file pattern applied correctly: `director-mode.md` loaded only when director-mode flags are present; base context stays lean for standalone invocations.
- Intent file locking protocol (no silent mutations after confirmation, append-only revision sections, decisions.md logging) is correct state-machine design.
- The discipline/intent alignment separation — assertions vs. advisory framing — is the right architectural split and prevents false precision in the intent section.
- CLARIFY one-round cap prevents infinite loops; its presence as a design constraint is correct even if the escalation path after the cap needs tightening.
- Superseded plan doc preserved with a clear header and pointers to the implementation — good institutional knowledge trail.

---
- *Persona:* claude-config-reviewer, correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,165 @@
...
+"Bash(gh api *)",
+"Bash(gh pr diff *)",
+"Bash(gh pr view *)",
+"Read(~/.claude/commands/git/verify-business-logic/**)",
+"Read(~/.claude/skill-references/request-interaction-base.md)"
+```
+
+## Reference Files (always loaded)
+
+- `discipline-rules.md` — **Read first.** Canonical structural rules for discipline checks.
```
ahoym: **[claude-config-reviewer, correctness-reviewer]** These two files are consumed on every invocation (Steps 5 and 8) but listed under `## Reference Files (always loaded)` while using backtick notation — not `@` eager loading. Per project convention, `@` references eagerly load on invocation; backtick references are conditional read instructions requiring explicit action at the point of use.

This creates two downstream problems:
1. Step 5 says `(already loaded in reference files step)` — that's inaccurate for backtick references; the agent may skip reading `discipline-rules.md` entirely.
2. `prompt-template.md`'s `{{DISCIPLINE_RULES}}` placeholder implies literal substitution, but if the file is a reference-file (not injected), the placeholder is never filled.

**Fix (pick one):** Switch to `@./discipline-rules.md` and `@./prompt-template.md`, or rename the section to `## Reference Files (read at skill start)` and add explicit read instructions at the top of Steps 5 and 8.

---
- *Persona:* claude-config-reviewer, correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,165 @@
...
+## Steps
+
+### Step 1: Parse Arguments and Detect Mode
+
+Parse `$ARGUMENTS`:
+- Extract `<pr-number>` (required)
+- Check for `--intent-file <path>` and `--session-dir <path>`
+- If both present: `MODE=director`, read `director-mode.md`
+- If neither present: `MODE=standalone`
+- If only one present: error — both flags required for director mode
```
ahoym: **[correctness-reviewer]** The `if only one present: error` branch is underspecified. "Error" could mean: halt and print a message, exit with non-zero, continue silently as standalone. In director mode, the caller is a background agent — a silent fallthrough to standalone behavior would corrupt session state (wrong mode, wrong output location, wrong confidence framing) with no signal to the director.

Spell out the error contract: (1) emit a clearly formatted error message, (2) halt before fetching any PR data, (3) return no output that could be mistaken for a successful result.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Step 2: Set Persona
+
+Check if `convergence-verifier` persona is active. If not, activate it:
+```
+/set-persona convergence-verifier
```
ahoym: **[claude-config-reviewer, correctness-reviewer]** Two issues with mid-skill persona activation:

1. **Ordering**: Step 1 runs before persona is active — argument parsing happens without the convergence-verifier lens, which affects interpretation of ambiguous input.
2. **Programmatic invocation**: In director mode, `/set-persona` is a slash-command-style invocation. If the Skill tool isn't available or the call is a no-op in headless context, the persona is never set — confidence framing and output tone degrade silently.

Recommendation: Either move persona activation to a **Step 0** (before Step 1) so the full skill runs with the persona active, or declare it in Prerequisites so operators activate before invoking. If keeping it in-step, use `Skill tool: skill="set-persona", args="convergence-verifier"` to ensure it works in both modes.

---
- *Persona:* claude-config-reviewer, correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,165 @@
...
+- PR reviews: all review submissions
+
+If PR is merged or closed, note terminal state but proceed — verification of completed PRs is valid.
+
+### Step 4: Resolve Intent
+
+**Standalone mode** (`MODE=standalone`):
+1. Read PR description and linked issue bodies
+2. Draft a quick intent summary following the format:
+   ```
```
ahoym, line 77: **[correctness-reviewer]** Step 4 director branch has no error handling for three reachable failure modes:
- Intent file path doesn't exist
- File exists but has no `Source` field
- `Source` value is unrecognized (not in `director-negotiated | agent-prompt | inferred-from-pr-description`)

All three silently fall through to Step 6's confidence-framing logic with undefined `INTENT_SOURCE`. An unrecognized Source silently produces the wrong confidence framing on the verification report — which is exactly the output the verifier is supposed to guarantee is correct.

Also: consider validating the intent file before Step 3's PR data fetch. A malformed intent file wastes the full PR fetch (diff, comments, reviews) and leaves no partial state to recover from.

Explicit error cases: halt on each, emit a descriptive message, return no output the director could parse as success.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,165 @@
...
+3. Failed assertions: cite specific comment IDs, thread URLs, or content
+
+Build the discipline checklist per the format in `discipline-rules.md`.
+
+### Step 6: Run Intent Alignment Checks
+
+Using the resolved intent and the PR diff/comments:
+
+1. **Acceptance criteria**: For each criterion, search the diff and commits for evidence of delivery. Cite commit SHA + diff lines.
+2. **Scope analysis**: Identify changes not covered by any acceptance criterion. Classify each as:
```
ahoym, line 118: **[correctness-reviewer]** CLARIFY is triggered at Step 7 — after discipline checks (Step 5) and partial intent alignment (Step 6) are already complete. On re-invocation the verifier re-fetches PR state from scratch. If new comments were posted during the CLARIFY round-trip, the second run's discipline findings reflect a different PR snapshot than the first run's discipline checks did — the report silently mixes findings from two different states.

The fix is to check intent vagueness earlier — ideally at Step 4 after reading the intent but before fetching PR data. If intent is too vague to evaluate, CLARIFY immediately; discipline checks (which don't need intent) can still run on the re-invocation without the snapshot inconsistency.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/director-mode.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Success signals
+<observable outputs the verifier can check at convergence>
+```
+
+**Source field affects confidence**:
+- `director-negotiated`: highest confidence — operator confirmed. Full scope-drift analysis warranted.
+- `agent-prompt`: medium — restructured from the agent's prompt, not directly confirmed by operator. Note lower confidence in report.
+- `inferred-from-pr-description`: lowest — tone down scope-drift checks. Frame as suggestions, not assertions.
+
```
ahoym, line 35: **[correctness-reviewer]** When the verifier hits the one-CLARIFY limit and produces an 'intent too vague to evaluate' section, the director only logs `verifier-clarification` to `decisions.md`. The operator sees this buried in the Phase 5 summary — there is no mechanism surfacing it as a blocking finding. The session can converge with an unverified PR.

Define a clear escalation path for the post-limit case: either (a) the session explicitly cannot converge without a satisfactory intent verification (the verifier's degraded output blocks the Phase 5 summary from showing "all PRs verified"), or (b) the degraded report is labeled `UNVERIFIED` in a machine-readable way so the director can escalate rather than log-and-continue.

Related: the Step 9 completion signal (`<pass/fail summary>`) is free-text with no defined structure for machine parsing. A structured prefix like `VERIFIED:intent-unclear` would let the director act without free-text parsing.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/discipline-rules.md
```diff
@@ -0,0 +1,95 @@
...
+**Assertion**: For threads classified as resolved or acknowledged:
+- A reaction emoji exists (rocket for resolved, thumbsup for acknowledged)
+- No text reply from the reviewer *after* the addresser's fix/acknowledgement
+- Exception: operator comments (no `Role:` tag) always reopen the thread
+
+**How to check**: For each thread with a reviewer reaction emoji, verify no subsequent reviewer text reply exists on that thread. Ignore operator comments that reopen.
+
+**What to cite**: Thread ID + the unnecessary text reply comment ID.
+
+## Rule 4: Loop Closure
```
ahoym, line 44: **[correctness-reviewer]** Rule 3's operator-reopen exception (`operator comments (no Role: tag) always reopen the thread`) is undefined — what makes an operator comment a "reopen"? A new question, explicit disagreement, and "thanks!" all lack a `Role:` tag, but only some should reopen the thread.

If the verifier applies subjective judgment here, Rule 3 stops being a pass/fail assertion and becomes an opinion — which undermines the discipline section's claimed determinism. Define minimum conditions for "reopen": e.g., "operator comment that poses a new question, requests additional changes, or explicitly disagrees with the fix." Without this, the rule produces false positives that damage trust in the discipline section.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/discipline-rules.md
```diff
@@ -0,0 +1,95 @@
...
+**Assertion**: Agent-posted comments use backtick-wrapped `` `#N` `` or omit `#` entirely.
+
+**How to check**: Search agent comments for bare `#\d+` not inside backticks or code fences.
+
+**What to cite**: Comment ID + the bare reference.
+
+**Severity**: Low — cosmetic, but breaks when referencing non-existent issues.
+
+## Rule 6: Empty Review Guard
+
```
ahoym, line 73: **[correctness-reviewer]** Rule 4's timing caveat — "check timing — if the address runner converged and the thread is still orphaned, it's a finding" — is unevaluable in standalone mode. The verifier only sees PR comment timestamps; it has no access to runner `status.md` files or session state.

In standalone mode, the verifier cannot know whether the address runner converged. Specify the behavior explicitly:
- **Standalone**: Flag orphaned threads with the timing caveat in the report text ("address runner state unknown — verify before treating as discipline failure").
- **Director mode**: The Phase 5 ordering guarantees the verifier is called post-convergence; the timing caveat is unnecessary and should be skipped.

This makes Rule 4 evaluable in both modes without requiring unresolvable state.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
```diff
@@ -53,6 +53,19 @@ For prompt-free execution, add these allow patterns to `~/.claude/settings.local
    ```markdown
    # Director Decisions — <timestamp>
    ```
+8. **Capture intent** for each item. Create `<session_dir>/intents/` directory. For each PR/issue in scope:
```
ahoym, line 57: **[claude-config-reviewer]** Step 8b references the intent file schema via prose cross-reference: "following the schema in `/verify-business-logic`'s `director-mode.md`." This is a sibling-skill cross-reference — a drift risk. If the schema evolves in `director-mode.md` (e.g., new required fields, renamed sections), the director's intent capture won't automatically know; it will silently write intent files that don't match the verifier's expected format.

Consider extracting the intent schema to `skill-references/intent-file-schema.md` shared by both skills. At minimum, add a maintenance note: "Schema is defined and owned by `director-mode.md` — update this step if the schema changes."

---
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
```diff
@@ -53,6 +53,19 @@ For prompt-free execution, add these allow patterns to `~/.claude/settings.local
...
+   a. Read item metadata (PR title/description/linked issues, or issue body).
```
ahoym, line 58: **[correctness-reviewer]** Step 8c waits for operator confirmation with no timeout or fallback. The director's primary use case is background/automated sessions where operator availability isn't guaranteed. Operator non-response means the intent capture step stalls indefinitely — the session cannot proceed to Phase 2 (assess + generate artifacts) until intents are confirmed for all items.

Define a fallback: after a reasonable wait period, proceed with `Source: inferred-from-pr-description` and log the fallback to `decisions.md`. The director's existing Decision Framework has an analog for routine autonomous decisions — intent capture timeout is the same pattern: decide autonomously at lower confidence, surface what was decided.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
```diff
@@ -129,7 +146,14 @@ Note: process lifecycle (inactivity detection, kill, retry) is owned by the runn
 
 1. Check convergence by reading `*/status.md` across all run_dirs listed in `session.json`.
 2. Session ends when all runs converge.
-3. Write final `director-state.md` per playbook format with terminal state.
-4. Present a final summary: per-run retro (read each run's `*/results.md` and `*/learnings.md`).
-5. Review `<session_dir>/decisions.md` as part of the retro — surface decide-with-report entries (dissent, cost-time, out-of-scope) so the operator can audit autonomous calls and flag any that should have been escalated.
-6. Offer to invoke `/session-retro`. When accepted, the retro skill will read all worker `learnings.md` files before compounding — high-value worker observations are easy to lose otherwise.
+3. **Run convergence verifier** for each converged PR:
```
ahoym, line 166: **[claude-config-reviewer, correctness-reviewer]** Two issues with this invocation:

1. **Slash command in headless context**: The director runs via `claude -p` (headless). Phase 2 correctly uses `Skill tool: skill="sweep:review-prs"` — Phase 5 should match: `Skill tool: skill="git:verify-business-logic", args="<pr-number> --intent-file <path> --session-dir <path>"`.

2. **Naming mapping gap**: The path uses `<id>` but Phase 1 documents the convention as `pr-<N>.md` (e.g., `pr-78.md`) while the verifier output uses `verify-pr-<N>.md`. No explicit rule converts `<pr-number>` (integer) to `<id>` (prefixed string). If the director passes `78.md` instead of `pr-78.md`, the verifier hits an unhandled file-not-found error. Specify the mapping rule explicitly here.

---
- *Persona:* claude-config-reviewer, correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Good catch — the loaded learnings on this topic (`skill-references-and-loading.md` § "Eager-Load References That Are Always Consumed") confirm: when a reference is consumed on every invocation, `@` eager loading saves the tool call and guarantees availability. The context cost is identical either way.

Going with option 1: switching to `@./discipline-rules.md` and `@./prompt-template.md`. This also makes Step 5's "(already loaded in reference files step)" accurate instead of aspirational.

The `{{DISCIPLINE_RULES}}` placeholder in `prompt-template.md` is conceptual (agent fills it when composing the report), so eager-loading both files makes the agent's job clearer — both are in context when it hits the template.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed — silent fallthrough to standalone in a director-called context is the worst failure mode. I'll spell out the error contract: emit a descriptive message naming the missing flag, halt before any PR fetch, and produce no output that could be parsed as success.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed on both points.

1. **Ordering**: Step 1 is mechanical argument parsing — the persona lens doesn't change parsing behavior. But moving persona earlier is still correct form: it ensures the verifier lens is active for all subsequent judgment calls, including any edge cases in mode detection.

2. **Headless invocation**: The Skill tool syntax is the right fix. Slash commands don't resolve in `claude -p` sessions.

I'll promote persona activation to Step 0 (before argument parsing) and use `Skill tool: skill="set-persona", args="convergence-verifier"` syntax.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed on all three failure modes and the early validation suggestion. An unrecognized `Source` silently producing wrong confidence framing is exactly the kind of silent corruption the verifier is supposed to prevent.

I'll add explicit error handling for: (1) missing intent file, (2) absent `Source` field, (3) unrecognized `Source` enum value. Each halts with a descriptive message and produces no output the director could parse as success.

I'll also move intent file validation before the PR data fetch — a malformed intent file should be caught before burning API calls on diff/comments/reviews.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Good catch on the snapshot inconsistency. The key insight: discipline checks don't need intent (they're structural assertions), but intent alignment does. If intent is vague, CLARIFY should fire before PR data is fetched — not after discipline checks have already run on a snapshot that may be stale by re-invocation.

I'll restructure the flow for director mode:
- Step 4 reads the intent file (moved before Step 3)
- Vagueness check happens immediately after reading intent
- If vague → CLARIFY and stop (no PR data fetched yet)
- If clear → Step 3 fetches PR data → Steps 5-6 run on a single consistent snapshot

For standalone mode, the interactive capture in Step 4 already handles vagueness inline (operator confirms or revises), so the restructuring mainly affects director mode.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/director-mode.md
ahoym: Agreed — the log-and-continue path is the silent failure mode. A degraded "intent too vague" report should block convergence, not just document it.

I'll add:
1. A structured completion signal: `VERIFIED:<status>` prefix (e.g., `VERIFIED:pass`, `VERIFIED:fail`, `VERIFIED:intent-unclear`). The director can parse this without free-text matching.
2. Update the CLARIFY protocol's post-limit behavior: the degraded report includes a machine-readable `UNVERIFIED` marker that the director treats as blocking — the session cannot converge on an unverified PR.
3. Update Step 9 to use the structured prefix format.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/discipline-rules.md
ahoym: Agreed — without specific criteria, Rule 3 becomes subjective where it claims determinism. I'll add explicit conditions for what constitutes a "reopen":

An operator comment reopens the thread when it: (a) poses a new question, (b) requests additional changes, or (c) explicitly disagrees with the resolution. Comments that don't reopen: acknowledgments ("thanks"), praise, or tangential discussion unrelated to the thread's finding.

This keeps the rule evaluable as a pass/fail assertion rather than a judgment call.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/discipline-rules.md
ahoym: Agreed — the timing caveat is genuinely unevaluable without runner state. I'll add explicit mode-specific behavior:

- **Standalone**: Flag orphaned threads with a caveat: "address runner state unknown — verify manually before treating as discipline failure." This prevents false positives while still surfacing the finding.
- **Director mode**: Phase 5 ordering guarantees the verifier runs post-convergence. Skip the timing caveat — an orphaned thread at that point is a real finding.

This makes Rule 4 evaluable in both modes without requiring cross-session state access.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Agreed on the drift risk — the director writes intent files using the schema and the verifier reads them, so there are two consumers already.

However, extracting to `skill-references/intent-file-schema.md` expands this PR's scope. For now I'll add a maintenance note at the prose cross-reference: "Schema is defined and owned by `verify-business-logic/director-mode.md` — update this step if the schema changes." This makes the coupling explicit and auditable.

The shared reference extraction is a good follow-up — I'll note it for a separate PR when either skill next evolves the schema.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Agreed — operator non-response is a real deadlock risk in background sessions. I'll add a fallback: after 5 minutes of no response, proceed with `Source: inferred-from-pr-description` and log the fallback to `decisions.md` with category `intent-capture-timeout`.

This aligns with the Decision Framework's existing pattern for routine autonomous decisions: decide at lower confidence, surface what was decided.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Agreed on both sub-points:

1. **Slash command syntax**: Phase 2 uses `Skill tool: skill="sweep:review-prs"` — Phase 5 should match: `Skill tool: skill="git:verify-business-logic", args="<pr-number> --intent-file <path> --session-dir <path>"`. The `/verify-business-logic` syntax won't resolve in `claude -p`.

2. **Naming mapping**: I'll add an explicit mapping rule. The convention from Phase 1 Step 8d is `<id>` = `pr-<N>` (e.g., PR 78 → `pr-78.md`). I'll make this explicit at the Phase 5 invocation point so there's no ambiguity.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 7b4044f

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 7b4044f

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 7b4044f

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 7b4044f

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 7b4044f

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/director-mode.md
ahoym: Fixed in 7b4044f

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/discipline-rules.md
ahoym: Fixed in 7b4044f

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/discipline-rules.md
ahoym: Fixed in 7b4044f

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Fixed in 0b50066

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Fixed in 0b50066

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Fixed in 0b50066

---
- *Persona:* claude-config-expert
- *Role:* Addresser

ahoym:
11 suggestions implemented — see inline comments.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

ahoym (commented):
## Team Re-review: feat: add standalone verify-business-logic skill

Both fix commits (`7b4044f`, `0b50066`) comprehensively address all HIGH and MEDIUM findings from cycle 1. The updated SKILL.md is substantially better: intent validation runs before PR data fetch, CLARIFY fires early, persona activation is Step 0 with Skill tool invocation, and the structured `VERIFIED:<status>` completion signal makes the director/verifier contract machine-readable.

Reviewers (this cycle): claude-config-reviewer, correctness-reviewer
Carried forward: n/a (all domains affected by fixes)

### Previous Findings

- ✅ 11 resolved (code change verified)
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

None from the new commits. A few LOW items noted in the cycle 1 summary remain unaddressed (`prompt-template.md` `{{DISCIPLINE_RULES}}` placeholder still inconsistent with eager-loaded reference, `## Context` block still present but unused, `## Usage` section duplicates frontmatter) — these are cosmetic and don't block correctness.

### Positive Signals

- Intent validation moved to Step 2 (before PR fetch) eliminates the TOCTOU risk and makes errors bounded — a clean structural fix, not just a safeguard bolted on.
- The `VERIFIED:intent-unclear` blocking signal closes the session-convergence gap cleanly: the director can now gate on verifier status without free-text parsing.
- Rule 3 and Rule 4 in `discipline-rules.md` are now deterministic assertions: reopen criteria are enumerated, and timing caveats are mode-specific. The discipline section can now genuinely claim pass/fail without subjective judgment.
- The 5-minute intent-capture timeout with `inferred-from-pr-description` fallback is the right pattern: autonomous at low confidence, logged, never deadlocking.

---
- *Persona:* claude-config-reviewer, correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
```diff
@@ -148,10 +163,17 @@ Note: process lifecycle (inactivity detection, kill, retry) is owned by the runn
...
+   a. Invoke via Skill tool: `skill="git:verify-business-logic", args="<pr-number> --intent-file <session_dir>/intents/<id>.md --session-dir <session_dir>"`. **Naming mapping**: `<id>` uses the `pr-<N>` convention from Phase 1 Step 8d (e.g., PR 78 → `pr-78.md`).
+   b. If the verifier outputs `CLARIFY: <question>`, route through the Decision Framework:
```
ahoym: **[CRITICAL] [correctness-reviewer]** Phase 5 Step 3b has no exit condition — infinite CLARIFY loop risk

Step 3b: if verifier outputs `CLARIFY:`, route through Decision Framework, re-invoke verifier. If the re-invoked verifier outputs another `CLARIFY:` — which it can, because the verifier is a fresh Skill instance with zero memory of prior runs — Step 3b loops again with no counter, no fallback, no terminal condition.

Worst-case: vague intent → CLARIFY → director answers silently (simple/silent route, no operator escalation) → fresh re-invoke → same vagueness → CLARIFY → repeat. Spins invisibly.

Root cause: the one-CLARIFY rule has no protocol enforcement across independent invocations (see companion finding in `verify-business-logic/SKILL.md` Important Notes).

**Fix:** Add exit condition to Step 3b: "If re-invoked verifier outputs a second `CLARIFY:`, treat as `VERIFIED:intent-unclear`, log `verifier-clarification-loop` to `decisions.md`, do not re-invoke." Pairs with `--clarify-answered` flag on the verifier side.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
```diff
@@ -148,10 +163,17 @@ Note: process lifecycle (inactivity detection, kill, retry) is owned by the runn
...
+      - Simple/silent: answer from context and re-invoke verifier.
+      - Complex/escalate: ask operator, then re-invoke verifier with the answer.
+      - Log to `decisions.md` with category `verifier-clarification`.
+   c. Verifier posts a top-level PR comment and writes `<session_dir>/verify-pr-<N>.md`.
```
ahoym: **[HIGH] [correctness-reviewer]** No handler for verifier error output — only `CLARIFY:` and implicit `VERIFIED:*` are handled

Verifier SKILL.md Step 2 explicitly halts on three error conditions and produces "no output the director could parse as success." But director Phase 5 only handles two output shapes: `CLARIFY:` (Step 3b) and (implicitly) `VERIFIED:*` (proceeds to Step 4).

If the verifier halts with an error, the director receives error text matching neither. Most likely outcome: silently treats invocation as done and moves to Step 4, omitting verification.

**Fix:** Add Step 3c: "If verifier output is neither `CLARIFY:` nor `VERIFIED:*`, treat as verifier error. Log to `decisions.md` as `verifier-error`. Do not proceed to Step 4 for this PR without operator decision — either retry with corrected args or mark as `verification-skipped` with explicit acknowledgment."

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
```diff
@@ -148,10 +163,17 @@ Note: process lifecycle (inactivity detection, kill, retry) is owned by the runn
 
 1. Check convergence by reading `*/status.md` across all run_dirs listed in `session.json`.
 2. Session ends when all runs converge.
-3. Write final `director-state.md` per playbook format with terminal state.
-4. Present a final summary: per-run retro (read each run's `*/results.md` and `*/learnings.md`).
-5. Review `<session_dir>/decisions.md` as part of the retro — surface decide-with-report entries (dissent, cost-time, out-of-scope) so the operator can audit autonomous calls and flag any that should have been escalated.
-6. Offer to invoke `/session-retro`. When accepted, the retro skill will read all worker `learnings.md` files before compounding — high-value worker observations are easy to lose otherwise.
+3. **Run convergence verifier** for each converged PR:
```
ahoym, line 166: **[HIGH] [correctness-reviewer]** Phase 5 invokes verifier for all converged PRs, but mid-session additions may lack intent files

Phase 1 Step 8 captures intent for PRs in scope at session start. The lock-and-update cycle appends to existing intent files — it does not create new ones for PRs added mid-session. If such a PR converges, Phase 5 invokes `--intent-file <session_dir>/intents/<id>.md` pointing to a file that does not exist → verifier halts with error → director has no error handler (see companion finding).

**Fix:** Before invoking verifier for a PR in Phase 5, check whether `<session_dir>/intents/<id>.md` exists. If missing: (a) run Phase 1 Step 8 intent capture for this PR first, OR (b) invoke verifier in standalone mode and log the gap to `decisions.md`.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,174 @@
...
+
+The `intent-unclear` status is a **blocking signal** — the director must not converge the session on a PR with this status.
+
+## Important Notes
+
+- **Discipline checks are assertions, not opinions.** They pass or fail with cited evidence. Don't soften discipline findings.
+- **Intent alignment is advisory.** Frame as "things to check," not verdicts. The operator decides what matters.
+- **Don't repeat the diff.** The operator has the PR open. Cite specific locations, don't quote large blocks.
+- **Footnote is mandatory.** The `Role: Verifier` tag distinguishes verification comments from review/address comments.
+- **One CLARIFY max in director mode.** Checked at Step 2 before PR data is fetched. If still unclear after one round, produce the report with explicit "couldn't evaluate" sections and `VERIFIED:intent-unclear` status rather than looping.
```
ahoym: **[HIGH] [claude-config-reviewer, correctness-reviewer]** One-CLARIFY-per-run rule has no enforcement mechanism across fresh re-invocations

"Maximum one CLARIFY per verifier run" is stated here (Important Notes) and in `director-mode.md`. But when the director re-invokes the verifier as a fresh Skill tool call, the new instance has zero memory of prior runs. Nothing in the intent file, args, or session dir signals "this is attempt 2." The new instance reads the same vague intent file, hits the same vagueness at Step 2, and can legally emit another `CLARIFY:`.

This is the protocol-level root cause of the infinite-loop risk in director Phase 5 Step 3b.

**Fix:** Add a `--clarify-answered` flag the director passes on re-invocation. In Step 2, when this flag is present and intent is still vague, skip CLARIFY and emit `VERIFIED:intent-unclear` directly. Document in both SKILL.md Step 2 and `director-mode.md`.

---
- *Persona:* claude-config-reviewer, correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,174 @@
...
+
+   ## Acceptance criteria
+   - [ ] <criterion — checkable>
+
+   ## Out of scope
+   - <inferred exclusions, if any>
+   ```
+2. Present to operator: "Here's what I think this PR is trying to do — confirm or revise?"
+3. One revision round if operator adjusts
+4. If still too vague after one round, produce report with "intent too vague to evaluate" section
```
ahoym: **[HIGH] [correctness-reviewer]** Standalone mode missing `VERIFIED:intent-unclear` exit status on too-vague path

Step 2 standalone says "produce report with 'intent too vague to evaluate' section" after one failed revision round. But Step 7 defines `VERIFIED:intent-unclear` as a **blocking signal** — the director must not converge on a PR with this status.

Step 7's description anchors `intent-unclear` to "after CLARIFY limit reached" — but CLARIFY is director-mode-only. Standalone's terminal state is underdefined: a report is produced but no machine-readable exit status.

**Fix:** Add `VERIFIED:intent-unclear` to standalone Step 2's too-vague path. Update Step 7 description to generalize: "emitted when intent is too vague to evaluate, regardless of mode."

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/prompt-template.md
```diff
@@ -0,0 +1,75 @@
...
+## Input Placeholders
+
+| Placeholder | Source | Required |
+|-------------|--------|----------|
+| `{{INTENT}}` | Intent file content (director mode) or self-captured summary (standalone) | Yes |
+| `{{INTENT_SOURCE}}` | `director-negotiated`, `operator-confirmed`, or `inferred-from-pr-description` | Yes |
+| `{{DIFF}}` | Full PR diff from `gh pr diff` | Yes |
+| `{{COMMENTS}}` | All PR comments and review threads from `gh` API | Yes |
+| `{{DISCIPLINE_RULES}}` | Content of `discipline-rules.md` | Yes |
+| `{{MODE}}` | `standalone` or `director` | Yes |
```
ahoym, line 13: **[HIGH/MEDIUM] [claude-config-reviewer, correctness-reviewer]** `DISCIPLINE_RULES` placeholder is vestigial — listed as Required but never substituted

The Input Placeholders table marks `DISCIPLINE_RULES` as Required, sourced from "Content of `discipline-rules.md`". But: (1) `discipline-rules.md` is already eager-loaded via `@./discipline-rules.md` in SKILL.md — content is in context automatically, no substitution needed; (2) the Report Template body contains no `DISCIPLINE_RULES` insertion point — the Discipline Check section has hardcoded structure; (3) no step in SKILL.md says "populate `DISCIPLINE_RULES`".

A Required placeholder with no substitution target and no population step is a contradiction.

**Fix:** Remove `DISCIPLINE_RULES` from the Input Placeholders table entirely.

---
- *Persona:* claude-config-reviewer, correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/set-persona/convergence-verifier.md
```diff
@@ -0,0 +1,21 @@
+# Convergence Verifier
```
ahoym, line 1: **[HIGH] [claude-config-reviewer]** Persona missing `## Extends: reviewer`, `## Proactive Cross-Refs`, and `code-quality-instincts.md` reference

Three structural conventions missing that every other reviewer persona in this repo follows:

1. **`## Extends: reviewer`** — From `skill-evolution.md`: all domain reviewer personas extend the base `reviewer`. This persona does not, missing inherited baseline quality instincts and proactive cross-refs.

2. **`## Proactive Cross-Refs`** — Per `personas.md`: every persona domain gets this section. CLARIFY protocol rules, timing caveats, and intent source confidence scaling are gotchas that belong in a `convergence-verifier-gotchas.md` companion, not scattered inline.

3. **`code-quality-instincts.md` reference** — All reviewer personas reference this shared learning. This one does not.

**Fix:** Add `## Extends: reviewer`, create a companion gotchas file, add `## Proactive Cross-Refs`.

---
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/director-mode.md
```diff
@@ -0,0 +1,70 @@
...
+## Acceptance criteria
+- [ ] <criterion 1 — checkable>
+- [ ] <criterion 2 — checkable>
+
+## Out of scope
+- <thing the operator explicitly didn't want pulled in>
+
+## Success signals
+<observable outputs the verifier can check at convergence>
```
ahoym, line 27: **[MEDIUM] [correctness-reviewer]** One-CLARIFY constraint documented here but has no enforcement mechanism

"Maximum one CLARIFY per verifier run" is this file's authoritative rule. But the verifier is re-invoked as a fresh Skill instance with no memory of prior runs. Nothing in args, intent file, or session dir signals "attempt 2." The constraint relies on LLM behavioral consistency across independent invocations — not a protocol-level guarantee.

This file owns the CLARIFY spec and is the right place to define enforcement.

**Fix:** Document enforcement here alongside the rule: a `--clarify-answered` flag the director passes on re-invocation. State explicitly that Step 2 must check for this flag and skip to `VERIFIED:intent-unclear` if present and intent still vague.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

ahoym (commented):
## Team Review: feat: add standalone verify-business-logic skill

Solid standalone-first design with clean mode separation, explicit error halts, and a well-scoped CLARIFY protocol. The main themes to address before merge: (1) a concrete infinite-loop risk in director Phase 5 where the one-CLARIFY rule has no enforcement mechanism across fresh Skill re-invocations, (2) missing error handlers in Phase 5 — no handler for verifier error output and no check for missing intent files on mid-session PRs, and (3) a cluster of always-on token costs that should be trimmed or converted to conditional loading.

Reviewed by: claude-config-reviewer, correctness-reviewer

### Findings

16 finding(s) — see inline comments.

### Positive Signals

- Standalone-first design is correctly implemented: standalone gathers intent interactively, director mode accepts pre-computed intent files and skips gathering — exactly the right composability pattern. [claude-config-reviewer]
- Explicit halt-with-error in Step 2 director mode (file not found, missing Source, unrecognized Source value) produces no parseable success output on error — prevents silent state corruption downstream. [correctness-reviewer]
- CLARIFY check placed at Step 2 before PR data fetch correctly minimizes wasted API calls on ambiguous intents. [correctness-reviewer]
- Intent file locking semantics are clean: written once, never silently mutated, revision history append-only with dated sections and incremented counter. [correctness-reviewer]
- Discipline rules correctly framed as assertions (pass/fail with cited evidence) distinct from advisory intent alignment — well-maintained across all four files. [claude-config-reviewer]
- Rule 4 (Loop Closure) timing caveat correctly differentiates standalone vs director mode, avoiding false positives while treating orphaned threads as real findings when Phase 5 ordering guarantees post-convergence execution. [correctness-reviewer]
- Structured `VERIFIED:<status>` completion signal lets the director parse status without free-text matching. [correctness-reviewer]
- Pre-filter for converged items on director relaunch cycles (Phase 2) avoids wasted session startup cost. [claude-config-reviewer]

---
- *Persona:* claude-config-reviewer, correctness-reviewer
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
```diff
@@ -54,6 +54,19 @@ For prompt-free execution, add these allow patterns to `~/.claude/settings.local
    ```markdown
    # Director Decisions — <timestamp>
    ```
+8. **Capture intent** for each item. Create `<session_dir>/intents/` directory. For each PR/issue in scope:
+   a. Read item metadata (PR title/description/linked issues, or issue body).
+   b. Draft an intent file from metadata following the schema in `/verify-business-logic`'s `director-mode.md`. (**Maintenance note:** Schema is defined and owned by `verify-business-logic/director-mode.md` — update this step if the schema changes.)
```
ahoym: **[MEDIUM] [claude-config-reviewer]** Schema cross-reference path `/verify-business-logic` is not a resolvable file path

Step 8b says "following the schema in `/verify-business-logic`'s `director-mode.md`" — `/verify-business-logic` is not a file system path the director can `Read` at runtime. If the director reads the schema at intent-capture time (which the maintenance note implies it should), it needs the full path.

**Fix:** Change to `~/.claude/commands/git/verify-business-logic/director-mode.md`. Add explicit `Read` instruction at Phase 1 Step 8b: "Read `~/.claude/commands/git/verify-business-logic/director-mode.md` for the current intent file schema before drafting."

---
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/prompt-template.md
```diff
@@ -0,0 +1,75 @@
...
+
+| Criterion | Delivered? | Evidence |
+|-----------|-----------|----------|
+| <criterion from intent> | Yes/No/Partial | <commit SHA, diff lines, or "not found"> |
+
+#### Scope
+
+- **In scope, delivered**: <items from intent that landed>
+- **Intentional expansion**: <items pulled in beyond intent, with justification visible in PR>
+- **Unrelated drift**: <changes unrelated to intent — surface, don't judge>
```
ahoym, line 43: **[MEDIUM] [claude-config-reviewer]** Meta-comment inside template body will appear verbatim in posted PR comments

The line `> (uses INTENT_SOURCE as a dynamic placeholder in report)` is an authoring note placed inside the fenced template block. SKILL.md Step 6 says "Build the report following the template in `prompt-template.md`." If the agent uses this template as a starting point, this meta-comment appears in the actual PR comment posted via `gh api`.

**Fix:** Remove from inside the template block. Place in a `## Notes` section outside the fenced template if needed.

---
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,174 @@
+---
+name: verify-business-logic
+description: "Post-convergence verification — checks discipline rules and intent alignment on a PR after review/address cycles complete. Standalone or director-called."
+argument-hint: "<pr-number> [--intent-file <path> --session-dir <path>]"
+---
+
+## Context
```
ahoym: **[MEDIUM] [claude-config-reviewer]** `## Context` block values are never referenced — dead context cost on every invocation

The `git rev-parse` and `git branch` values are injected on every invocation but referenced in zero steps. This skill fetches PR data via `gh` using the `<pr-number>` argument — git context is irrelevant.

**Fix:** Remove the `## Context` block (lines 7-10).

---
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,174 @@
...
+"Bash(gh api *)",
+"Bash(gh pr diff *)",
+"Bash(gh pr view *)",
+"Read(~/.claude/commands/git/verify-business-logic/**)",
+"Read(~/.claude/skill-references/request-interaction-base.md)"
+```
+
+## Reference Files (always loaded)
+
+@./discipline-rules.md
```
ahoym: **[MEDIUM] [claude-config-reviewer]** `@./` prefix has known resolution issues — and both eagerly-loaded files exceed the 50-line guideline

From `skill-references-and-loading.md`: "`@./` relative paths may have resolution issues. Use `@filename.md` (skill-directory-relative without `./`) instead."

Beyond path syntax, both files exceed the less-than-50-line threshold for `@` eager loading: `discipline-rules.md` is 97 lines (consumed only at Step 4, skipped on CLARIFY early-exit) and `prompt-template.md` is 75 lines (consumed only at Step 6).

**Fix:** (1) Drop the `./`: use `@discipline-rules.md` and `@prompt-template.md`. (2) Consider converting both to conditional references read at Steps 4 and 6 — saves ~172 lines of context on CLARIFY early-exit paths.

---
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,174 @@
...
+@./prompt-template.md
+
+## Reference Files (conditional)
+
+- `director-mode.md` — **Read only when** `--intent-file` and `--session-dir` are present in `$ARGUMENTS`. Covers CLARIFY protocol, intent file schema, session dir output convention.
```
ahoym, line 34: **[LOW] [claude-config-reviewer]** Conditional reference `request-interaction-base.md` missing explicit `Read` instruction at Step 6

From `skill-references-and-loading.md`: include an explicit `Read` instruction in the step that consumes a conditional reference, not just in the `## Reference Files (conditional)` header.

Step 6 says "per `request-interaction-base.md` format" but does not tell the agent to read the file there.

**Fix:** Add to Step 6: "Read `~/.claude/skill-references/request-interaction-base.md` for the footnote format."

---
- *Role:* Team-Reviewer

claude/commands/git/verify-business-logic/SKILL.md
```diff
@@ -0,0 +1,174 @@
...
+
+# Verify Business Logic
+
+Post-convergence check that verifies review/address discipline and intent delivery on a PR. Produces a top-level PR comment with findings.
+
+Two modes:
+- **Standalone**: `/verify-business-logic 42` — self-captures intent from PR metadata, asks operator to confirm
+- **Director-called**: `/verify-business-logic 42 --intent-file <path> --session-dir <path>` — reads locked intent file, skips capture
+
+## Usage
```
ahoym: **[LOW] [claude-config-reviewer]** `## Usage` section duplicates `argument-hint` frontmatter — always-loaded redundancy

The `argument-hint` frontmatter already declares the same two forms. Both load on every invocation.

**Fix:** Delete `## Usage` (lines 20-23). The frontmatter covers discovery; Step 1 covers mode parsing.

---
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
```diff
@@ -54,6 +54,19 @@ For prompt-free execution, add these allow patterns to `~/.claude/settings.local
...
+   c. Present the draft to the operator: "Here's what I think we're building. Confirm, revise, or fill the gaps."
+   d. Restructure the operator's response into the final intent shape and write to `<session_dir>/intents/<id>.md` (e.g., `pr-78.md`, `issue-56.md`). Set `Source: director-negotiated`.
+   e. One revision round if the operator wants to tighten anything. **Timeout fallback**: if no operator response within 5 minutes, proceed with `Source: inferred-from-pr-description` and log to `decisions.md` with category `intent-capture-timeout`. This aligns with the Decision Framework's pattern for routine autonomous decisions at lower confidence.
+   f. Once confirmed (or timed out with fallback), the file is **locked** for the session — no silent mutations.
+   g. Optionally write `<session_dir>/intents/index.md` — one line per item for session-at-a-glance.
+
```
ahoym, line 65: **[LOW] [correctness-reviewer]** Phase 1 timeout `Source: inferred-from-pr-description` misrepresents actual provenance

Step 8e uses `Source: inferred-from-pr-description` on timeout. But by then the director has drafted intent from metadata AND presented the draft to the operator (Steps 8b-c). The operator saw it, just did not reply.

`inferred-from-pr-description` implies the operator never saw the intent. The verifier will apply lowest-confidence framing to an intent the operator actually reviewed.

**Fix (minor):** Add `Source: draft-timeout` with confidence tier between `agent-prompt` and `inferred-from-pr-description`. Update `director-mode.md` and `prompt-template.md` confidence framing sections to match.

---
- *Persona:* correctness-reviewer
- *Role:* Team-Reviewer

claude/plans/convergence-verifier.md
```diff
@@ -0,0 +1,178 @@
+# Convergence Verifier
```
ahoym, line 1: **[LOW] [claude-config-reviewer]** 178-line SUPERSEDED plan doc — delete or truncate to pointer-only

A SUPERSEDED file committed with full historical content is dead weight. Git history already preserves this content.

**Options:** (1) Delete the file. (2) Keep only the 9-line SUPERSEDED header pointing to the implementation, remove the rest.

---
- *Role:* Team-Reviewer

claude/commands/director/SKILL.md
ahoym: Agreed — this is a real infinite-loop risk. A fresh Skill instance has zero memory of prior CLARIFY rounds, so the one-round cap is unenforceable without a protocol-level signal.

I'll add: (1) `--clarify-answered` flag the director passes on re-invocation, (2) exit condition in Step 3b — if re-invoked verifier outputs a second `CLARIFY:`, treat as `VERIFIED:intent-unclear`, log `verifier-clarification-loop` to `decisions.md`, do not re-invoke.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Agreed — the verifier's explicit error halts produce non-parseable output, but the director has no handler for that case. Silent proceed-to-Step-4 is exactly the wrong outcome.

I'll add Step 3c: if verifier output is neither `CLARIFY:` nor `VERIFIED:*`, treat as verifier error. Log to `decisions.md` as `verifier-error`, do not proceed to Step 4 without operator decision.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Agreed — Phase 1 Step 8 only captures intent for PRs in scope at session start. Mid-session additions that converge hit a missing intent file → verifier error → no handler (companion finding).

I'll add a pre-invocation check to Phase 5: before invoking verifier for a PR, check whether `<session_dir>/intents/<id>.md` exists. If missing, run Phase 1 Step 8 intent capture for that PR first. This is cleaner than falling back to standalone mode since the director has the session context to negotiate intent properly.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed — the one-CLARIFY rule is stated but not enforced. A fresh Skill instance can't know it's attempt 2 without explicit signaling.

I'll add: when `--clarify-answered` is present in `$ARGUMENTS` and Step 2 finds intent still vague, skip CLARIFY and emit `VERIFIED:intent-unclear` directly. This pairs with the director-side exit condition (companion finding on Phase 5 Step 3b).

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed — standalone's too-vague path produces a report but no machine-readable exit status. The `VERIFIED:intent-unclear` signal should be mode-independent.

I'll add `VERIFIED:intent-unclear` to standalone Step 2's too-vague path and update Step 7's description to generalize: "emitted when intent is too vague to evaluate, regardless of mode."

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/prompt-template.md
ahoym: Agreed — `DISCIPLINE_RULES` is a contradiction: marked Required with no substitution target, no population step, and the content is already eager-loaded via `@./discipline-rules.md`. The placeholder is vestigial from before eager loading was added.

Removing `DISCIPLINE_RULES` from the Input Placeholders table entirely.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/set-persona/convergence-verifier.md
ahoym: Agreed on two of three:

1. **`## Extends: reviewer`** — correct, this persona inherits reviewer baseline. Adding it.
2. **`code-quality-instincts.md` reference** — correct, every reviewer persona should reference this. Adding via `## Proactive Cross-Refs`.

Pushing back on the companion gotchas file: the CLARIFY protocol rules, timing caveats, and intent source scaling are already documented in `SKILL.md` and `director-mode.md` — the authoritative sources. A companion gotchas file would duplicate that content, creating drift risk. The persona's job is to set the lens; the skill owns the procedural details. Per claude-config-expert priorities: "single source of truth" and "fewer artifacts > more complete coverage."

I'll add `## Extends: reviewer` and `## Proactive Cross-Refs` with `code-quality-instincts.md`, but not the companion gotchas file.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/director-mode.md
ahoym: Agreed — this file owns the CLARIFY spec and is the right place to define enforcement. I'll document the `--clarify-answered` flag alongside the one-CLARIFY rule: the director passes this flag on re-invocation, and Step 2 checks for it, skipping to `VERIFIED:intent-unclear` if present and intent still vague.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Agreed — `/verify-business-logic` is a skill namespace, not a file path the director can `Read`. I'll change to the full path `~/.claude/commands/git/verify-business-logic/director-mode.md` and add an explicit `Read` instruction at Phase 1 Step 8b.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/prompt-template.md
ahoym: Agreed — the `> (uses INTENT_SOURCE as a dynamic placeholder in report)` line is inside the fenced template block and would appear verbatim in posted PR comments. Moving it to a `## Notes` section outside the template.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed — `git rev-parse` and `git branch` values are never referenced in any step. This skill fetches everything via `gh` using the `<pr-number>` argument. Removing the `## Context` block.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed on both points:

1. **`@./` → `@`**: Dropping the `./` per `skill-references-and-loading.md` guidance.
2. **Conditional loading**: Both files exceed the 50-line threshold and are only consumed at Steps 4 and 6 respectively. On CLARIFY early-exit, eager loading wastes ~172 tokens of context. Converting both to conditional references with explicit `Read` instructions at their consumption points.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed — Step 6 references `request-interaction-base.md` for the footnote format but doesn't include an explicit `Read` instruction. Adding `Read ~/.claude/skill-references/request-interaction-base.md for the footnote format` to Step 6.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Agreed — the `argument-hint` frontmatter already declares both usage forms. The `## Usage` section is redundant always-loaded content. Removing it.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Good catch — the operator did see the intent draft (Steps 8b-c), so `inferred-from-pr-description` understates the confidence tier. I'll add `Source: draft-timeout` with a confidence tier between `agent-prompt` and `inferred-from-pr-description`, and update `director-mode.md` and `prompt-template.md` confidence framing to include it.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/plans/convergence-verifier.md
ahoym: Agreed — 178 lines of SUPERSEDED content is dead weight when git history preserves it. Going with option 2: keep only the 9-line SUPERSEDED header with pointers to the implementation, remove the rest.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Fixed in 1cda38a

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Fixed in 1cda38a

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Fixed in 1cda38a

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 02fe19b

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 02fe19b

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/prompt-template.md
ahoym: Fixed in 02fe19b

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/set-persona/convergence-verifier.md
ahoym: Partially fixed in 02fe19b — added `## Extends: reviewer` and `## Proactive Cross-Refs` with `code-quality-instincts.md`. Pushed back on companion gotchas file (see initial reply).

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/director-mode.md
ahoym: Fixed in 02fe19b

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Fixed in 1cda38a

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/prompt-template.md
ahoym: The meta-comment text `> (uses INTENT_SOURCE as a dynamic placeholder in report)` is not present in `prompt-template.md` at commit `eb6ecec` — grep confirms no match. This may reference text that was removed in a prior cycle. No change needed.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 02fe19b

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 02fe19b

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 02fe19b

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/git/verify-business-logic/SKILL.md
ahoym: Fixed in 02fe19b

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/commands/director/SKILL.md
ahoym: Fixed in 1cda38a (director/SKILL.md) and 02fe19b (director-mode.md, prompt-template.md)

---
- *Persona:* claude-config-expert
- *Role:* Addresser

claude/plans/convergence-verifier.md
ahoym: Fixed in 02fe19b — truncated to pointer-only header.

---
- *Persona:* claude-config-expert
- *Role:* Addresser

Issue #83, opened 2026-04-12
Inline platform-specific commands into skills via `!` preprocessing + setup-time symlink


ahoym:
## Problem

Skills that need platform-specific CLI commands (gh vs glab) currently rely on external reference files (`skill-references/{github,gitlab}/*.md`) loaded at runtime via lazy `Read` instructions. Agents frequently skip these reads because they assess low information gain — they "know" `gh pr create` from training data and don't bother reading the project-specific version. This causes them to miss project conventions (body-file-to-tmp-dir patterns, absolute path requirements, jq filter file workarounds, GraphQL for GitLab inline comments, etc.) and improvise incorrect commands.

Stronger instruction language in SKILL.md files ("you MUST read this file") has not reliably fixed the problem. Empirically, inlined commands produce the most consistent agent behavior.

The challenge: this repo is shared across two machines — one GitHub, one GitLab. Statically inlining platform-specific commands into tracked SKILL.md files would cause merge conflicts when both machines push changes (e.g., learnings).

## Proposed Solution

Combine two mechanisms:

### 1. Setup-time platform symlink

`setup-claude.sh` creates a single symlink that resolves the platform once at install time:

```bash
~/.claude/platform-commands → dotfiles/claude/skill-references/{github,gitlab}/commands/
```

Platform is selected via interactive prompt on first run, persisted for re-runs.

### 2. Granular command files + `!` preprocessing

Break the existing cluster files (`pr-management.md`, `comment-interaction.md`, etc.) into atomic command files — one operation per file, 3-5 lines each. Skills `!`-inline them at point of use:

```markdown
9. **Create the review:**
!`cat ~/.claude/platform-commands/create-review`
```

The `!` syntax is CLI-level preprocessing — it runs the command and replaces the placeholder with the output before the agent sees the skill. The agent receives the literal `gh pr create ...` or `glab mr create ...` command as if it was always hardcoded in the SKILL.md.

### Agent reliability
- **No reference file to skip** — commands are physically present in the prompt via preprocessing
- **No platform conditional to evaluate** — CLI resolves it; agent sees a single-path skill
- **No improvisation opportunity** — the exact command with project-specific conventions is right there

### No merge conflicts
- Every tracked file is identical on both machines
- SKILL.md files contain `!`cat ~/.claude/platform-commands/...`` — same text everywhere
- Command files under `skill-references/{github,gitlab}/commands/` are tracked but each machine only edits its own platform's files through the symlink
[...]

PR #87, opened 2026-04-13
feat: extract .sh platform command files (#83 — phase 1)

ahoym:
## Summary

- Extracts atomic `.sh` command files from existing cluster `.md` files — one shell operation per file, matching filenames across platforms
- 25 GitHub + 25 GitLab command files under `skill-references/{github,gitlab}/commands/`
- Phase 1 of 4 for issue #83. Skills are refactored in phase 3 to consume these via `!`cat`` includes.

Relates to https://github.com/ahoym/dotfiles/issues/83

### DIFF
--- a/claude/skill-references/github/commands/batch-fetch-reviews.sh
+++ b/claude/skill-references/github/commands/batch-fetch-reviews.sh
@@ -1,0 +1,5 @@
+# Write jq filter to tmp/claude-artifacts/jq-filters/jq-filter.jq via Write tool first:
+#   .[] | {number, title, state, comments, user: .user.login, head_branch: .head.ref, requested_reviewers: [.requested_reviewers[].login], created_at: .created_at[:10], merged_at: (.merged_at // "n/a")[:10], body: (.body // "(none)")[:400]}
+# Then:
+gh api 'repos/{owner}/{repo}/pulls?state=all&sort=created&direction=asc&per_page=<SIZE>&page=<PAGE>' \
+  | jq -cf tmp/claude-artifacts/jq-filters/jq-filter.jq
--- a/claude/skill-references/github/commands/post-review-comments.sh
+++ b/claude/skill-references/github/commands/post-review-comments.sh
@@ -1,0 +1,11 @@
+# Write JSON payload to tmp/claude-artifacts/change-request-replies/review-<number>.json, then:
+gh api repos/{owner}/{repo}/pulls/<number>/reviews \
+  --input tmp/claude-artifacts/change-request-replies/review-<number>.json
+# Clean up:
+rm tmp/claude-artifacts/change-request-replies/review-<number>.json
+
+# Payload format (tmp/claude-artifacts/change-request-replies/review-<number>.json):
+# {"event": "COMMENT", "body": "...", "comments": [{"path": "file.md", "line": 42, "side": "RIGHT", "body": "..."}]}
+# event: COMMENT, APPROVE, or REQUEST_CHANGES
+# line: line number in final version (RIGHT side of diff)
+# side: always RIGHT for new-version comments
--- a/claude/skill-references/github/commands/post-review-comments.sh
+++ b/claude/skill-references/github/commands/post-review-comments.sh
@@ -1,7 +1,7 @@
 # Write JSON payload to tmp/claude-artifacts/change-request-replies/review-<number>.json, then:
 gh api repos/{owner}/{repo}/pulls/<number>/reviews \
   --input tmp/claude-artifacts/change-request-replies/review-<number>.json
-# Clean up:
+# Optional: clean up payload after posting
 rm tmp/claude-artifacts/change-request-replies/review-<number>.json
 
 # Payload format (tmp/claude-artifacts/change-request-replies/review-<number>.json):
--- a/claude/skill-references/gitlab/commands/post-issue-comment.sh
+++ b/claude/skill-references/gitlab/commands/post-issue-comment.sh
@@ -1,0 +1,2 @@
+# GitLab issue operations (v2 stub — not fully implemented)
+glab issue note <IID> --message-file <path>
--- a/claude/skill-references/gitlab/commands/post-issue-comment.sh
+++ b/claude/skill-references/gitlab/commands/post-issue-comment.sh
@@ -1,2 +1,2 @@
 # GitLab issue operations (v2 stub — not fully implemented)
-glab issue note <IID> --message-file <path>
+glab issue note <IID> -m "$(cat <path>)"
--- a/claude/skill-references/gitlab/commands/post-review-comments.sh
+++ b/claude/skill-references/gitlab/commands/post-review-comments.sh
@@ -9,6 +9,7 @@
 # For complex mutations, write query to .graphql file and use uppercase -F:
 #   glab api graphql -F query=@tmp/claude-artifacts/change-request-replies/gql-1.graphql
 # Lowercase -f query=@file does NOT read the file — passes literal "@file" string.
+# Alternatively, inline the mutation directly (lowercase -f with literal string):
 glab api graphql -f query='mutation {
   createDiffNote(input: {
     noteableId: "gid://gitlab/MergeRequest/<mr_id>",
--- a/claude/skill-references/gitlab/commands/post-review-comments.sh
+++ b/claude/skill-references/gitlab/commands/post-review-comments.sh
@@ -1,0 +1,29 @@
+# GitLab has no single review API. Post inline comments via GraphQL, then summary as top-level comment.
+
+# Step 1: Get SHAs and MR internal ID:
+# glab api projects/:id/merge_requests/<number>/versions | jq '.[0] | {base_commit_sha, head_commit_sha}'
+# glab mr view <number> --output json | jq '.id'
+# The MR id (not iid) is needed for GraphQL noteableId: "gid://gitlab/MergeRequest/<id>"
+
+# Step 2: Post each inline comment via GraphQL:
+# For complex mutations, write query to .graphql file and use uppercase -F:
+#   glab api graphql -F query=@tmp/claude-artifacts/change-request-replies/gql-1.graphql
+# Lowercase -f query=@file does NOT read the file — passes literal "@file" string.
+glab api graphql -f query='mutation {
+  createDiffNote(input: {
+    noteableId: "gid://gitlab/MergeRequest/<mr_id>",
+    body: "<escaped_body>",
+    position: {
+      baseSha: "<base_sha>",
+      headSha: "<head_sha>",
+      startSha: "<base_sha>",
+      paths: { oldPath: "<file_path>", newPath: "<file_path>" },
+      newLine: <line_number>
+    }
+  }) { note { id } errors }
+}'
+
+# Body escaping: use jq -Rs on a file for JSON-escaped string. Always jq, never python3.
+# Line number: must match exact line in new file (1-indexed). Verify against file content, not diff hunk arithmetic.
+# For new lines (+): omit oldLine. Context lines: both oldLine and newLine. Removed lines (-): only oldLine.
+# Step 3: Post review summary as top-level comment (see post-top-level-comment.sh).
--- a/claude/skill-references/github/commands/check-existing-review.sh
+++ b/claude/skill-references/github/commands/check-existing-review.sh
@@ -1,0 +1,1 @@
+gh pr list --head <branch-name>
--- a/claude/skill-references/github/commands/checkout-review.sh
+++ b/claude/skill-references/github/commands/checkout-review.sh
@@ -1,0 +1,2 @@
+gh pr checkout <number>
+git pull origin <headRefName>
--- a/claude/skill-references/github/commands/consolidated-fetch.sh
+++ b/claude/skill-references/github/commands/consolidated-fetch.sh
@@ -1,0 +1,3 @@
+# Fetch state + reviews + top-level comments in a single call.
+# No --jq (avoids quoted string permission prompts).
+gh pr view <number> --json state,reviews,comments,number,title,headRefName,baseRefName
--- a/claude/skill-references/github/commands/count-total-reviews.sh
+++ b/claude/skill-references/github/commands/count-total-reviews.sh
@@ -1,0 +1,1 @@
+gh api 'repos/{owner}/{repo}/pulls?state=all&per_page=1' -i 2>&1 | grep -i 'link:'
--- a/claude/skill-references/github/commands/create-review.sh
+++ b/claude/skill-references/github/commands/create-review.sh
@@ -1,0 +1,2 @@
+# Write body via Write tool to tmp/claude-artifacts/change-request-replies/pr-body.md, then:
+gh pr create --base <base-branch> --title "<title>" --body-file tmp/claude-artifacts/change-request-replies/pr-body.md
--- a/claude/skill-references/github/commands/fetch-activity-signals.sh
+++ b/claude/skill-references/github/commands/fetch-activity-signals.sh
@@ -1,0 +1,3 @@
+# Quick-exit check for polling — commits, reviews, state, and top-level comments in one call.
+# No --jq. Parse JSON response in agent logic.
+gh pr view <number> --json commits,reviews,state,comments
--- a/claude/skill-references/github/commands/fetch-inline-comments.sh
+++ b/claude/skill-references/github/commands/fetch-inline-comments.sh
@@ -1,0 +1,8 @@
+# Full fetch (--paginate to get all comments beyond default 30-per-page limit)
+gh api repos/{owner}/{repo}/pulls/<number>/comments --paginate --jq '.[] | {id, path, line, body, user: .user.login, created_at}'
+
+# Incremental fetch — write jq filter to file first:
+# 1. Write to tmp/claude-artifacts/jq-filters/jq-filter.jq via Write tool:
+#      .[] | select(.created_at > "<TS>") | {id, path, line, body, user: .user.login, created_at}
+# 2. Then (use piped jq -f instead of --jq):
+# gh api repos/{owner}/{repo}/pulls/<number>/comments --paginate | jq -f tmp/claude-artifacts/jq-filters/jq-filter.jq
--- a/claude/skill-references/github/commands/fetch-issue-comments.sh
+++ b/claude/skill-references/github/commands/fetch-issue-comments.sh
@@ -1,0 +1,8 @@
+# Full fetch (--paginate to get all comments beyond default 30-per-page limit)
+gh api repos/{owner}/{repo}/issues/<number>/comments --paginate --jq '.[] | {id, body, user: .user.login, created_at}'
+
+# Incremental fetch — write jq filter to file first:
+# 1. Write to tmp/claude-artifacts/jq-filters/jq-filter.jq via Write tool:
+#      .[] | select(.created_at > "<TS>") | {id, body, user: .user.login, created_at}
+# 2. Then (use piped jq -f instead of --jq):
+# gh api repos/{owner}/{repo}/issues/<number>/comments --paginate | jq -f tmp/claude-artifacts/jq-filters/jq-filter.jq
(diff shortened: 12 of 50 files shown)


ahoym (commented):
## Team Review: feat: extract .sh platform command files (#83 — phase 1)

Clean Phase 1 extraction — 50 atomic command files with consistent naming across both platforms. Inline documentation covers the key gotchas (no `--jq`, absolute paths for `gh`/`glab`, `-f` vs `-F` flag semantics). Two small items to verify before Phase 3 wires these up.

Reviewed by: claude-config-reviewer

### Findings

- Cleanup embedded in `post-review-comments.sh` (GitHub) — constrains Phase 3 consumers needing post-only primitive
- `glab issue note --message-file` flag unverified — standard `glab` uses `-m/--message`, not `--message-file`

### Positive Signals

- GitLab `post-review-comments.sh` documents GraphQL `-F query=@file` gotcha — prevents a high-friction Phase 3 failure
- Incremental fetch patterns route through Write-then-jq to avoid quoted-string permission prompts
- GitLab stubs clearly marked as incomplete — safe guardrail for Phase 3 authors
- Consistent 1-to-1 filename mapping across GitHub and GitLab (25 each) makes platform-switching predictable

---
- *Role:* Team-Reviewer

claude/skill-references/github/commands/post-review-comments.sh
```diff
@@ -0,0 +1,11 @@
+# Write JSON payload to tmp/claude-artifacts/change-request-replies/review-<number>.json, then:
+gh api repos/{owner}/{repo}/pulls/<number>/reviews \
+  --input tmp/claude-artifacts/change-request-replies/review-<number>.json
+# Clean up:
+rm tmp/claude-artifacts/change-request-replies/review-<number>.json
```
ahoym, line 5: **[claude-config-reviewer]** The `rm` cleanup is embedded in the same primitive as the `gh api` post. When Phase 3 skills consume this via `!cat`, they'll inherit cleanup as part of the post step — with no way to post-only and conditionally clean up on success.

If the intent is that cleanup always follows post, a comment explaining that would help Phase 3 authors. If post-only might be needed, consider a separate `cleanup-review-payload.sh` primitive.

---
- *Role:* Team-Reviewer

claude/skill-references/gitlab/commands/post-issue-comment.sh
```diff
@@ -0,0 +1,2 @@
+# GitLab issue operations (v2 stub — not fully implemented)
+glab issue note <IID> --message-file <path>
```
ahoym: **[claude-config-reviewer]** `glab issue note` uses `-m/--message` for inline text, not `--message-file`. Standard pattern is:
```sh
glab issue note <IID> --message-file <path>
```
This flag may not exist — `glab` typically accepts `--message`/`-m` for note content. Worth verifying before Phase 3 wires this up, since a silently incorrect stub could block issue-comment workflows.

If `--message-file` doesn't exist, the alternative is routing through a temp file and `$(cat ...)` (same pattern as `create-review.sh`), or using `glab issue note <IID> -m "$(cat <path>)"`.

---
- *Role:* Team-Reviewer

claude/skill-references/github/commands/post-review-comments.sh
ahoym: Good catch on the coupling. These command files are reference recipes (consumers select which lines to use), so the `rm` isn't mechanically inherited — but I agree that's not obvious without a comment.

I'll add a comment making the cleanup step visually distinct and explicitly optional. A separate `cleanup-review-payload.sh` for a single `rm` feels like over-engineering here — if Phase 3 surfaces a real need for post-only, we can extract then.

---
- *Persona:* claude-config-author
- *Role:* Addresser

claude/skill-references/gitlab/commands/post-issue-comment.sh
ahoym: You're right — `glab issue note` uses `-m/--message`, not `--message-file`. The stub has an incorrect flag.

Updating to `glab issue note <IID> -m "$(cat <path>)"` which matches the `$(cat ...)` pattern used elsewhere in the GitLab command files and avoids inventing a flag that doesn't exist.

---
- *Persona:* claude-config-author
- *Role:* Addresser

claude/skill-references/github/commands/post-review-comments.sh
ahoym: Fixed in 9243ff1

---
- *Persona:* claude-config-author
- *Role:* Addresser

claude/skill-references/gitlab/commands/post-issue-comment.sh
ahoym: Fixed in ca676a0

---
- *Persona:* claude-config-author
- *Role:* Addresser

ahoym (commented):
## Team Re-review: feat: extract .sh platform command files (#83 — phase 1)

Both findings addressed in 9243ff1 and ca676a0. Deprecation commit (4652439b) is clean — consistent frontmatter pattern across all 11 cluster files, minor inconsistency in body notes (only `pr-management.md` and `platform-detection.md` get `> **Note:**` blocks) but zero runtime impact on deprecated files. Reviewed `ca676a0` — no new findings.

Reviewers (this cycle): claude-config-reviewer
Carried forward: none

### Previous Findings

- ✅ 2 resolved (code changes verified in 9243ff1, ca676a0)
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

0 new findings.

### Positive Signals

Both fixes landed as minimal, well-targeted commits — no scope creep. The `-m "$(cat <path>)"` fix for the glab stub correctly mirrors the pattern used in other GitLab command files, maintaining consistency across the platform.

---
- *Role:* Team-Reviewer

claude/skill-references/github/commands/batch-fetch-reviews.sh
```diff
@@ -0,0 +1,5 @@
+# Write jq filter to tmp/claude-artifacts/jq-filters/jq-filter.jq via Write tool first:
+#   .[] | {number, title, state, comments, user: .user.login, head_branch: .head.ref, requested_reviewers: [.requested_reviewers[].login], created_at: .created_at[:10], merged_at: (.merged_at // "n/a")[:10], body: (.body // "(none)")[:400]}
```
ahoym, line 2: can we verify if something like this works? where comments here actually trigger agent behavior?

ahoym: Yes — this cycle was triggered by your comment. 🎉

The sweep loop watermarks the latest inline comment ID after each pass. When a new comment arrives (your comment `3070852651` ≠ previous watermark `3070779522`), the director relaunches the reviewer agent with the new activity. So posting a comment on any line in this PR will wake the agent up on the next poll interval.

---
- *Role:* Team-Reviewer

ahoym: no I meant where the comment in the bash script triggers a Write like is prescribed in these bash script comments

ahoym: Ah, different question — sorry for the misdirect. Yes, that's the design.

When a skill uses `!cat skill-references/github/commands/batch-fetch-reviews.sh`, the file contents get inlined verbatim into the skill's instruction text. The LLM then reads `# Write jq filter to tmp/claude-artifacts/jq-filters/jq-filter.jq via Write tool first:` as a literal instruction and will attempt to execute the Write tool call before running the `gh api` command beneath it.

Shell comments (`#`) are invisible to the shell but fully visible to the LLM reading the inlined text. The dual-purpose nature is intentional: the comment is human documentation AND a step directive for the agent. This is why the file is laid out in imperative sequence ("Write X first, then run Y") rather than just the raw command.

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: feat: extract .sh platform command files (#83 — phase 1)

PR was rebased — 4 commits → 3 (new HEADs: 2934ac1, dd2c641, acc3bf5). Both previously-reviewed fixes are preserved with identical content. The deprecation commit (4652439b — marking cluster `.md` files as deprecated) was dropped from the rebase. Reviewed `acc3bf5` — no new findings.

Reviewers (this cycle): claude-config-reviewer
Carried forward: none

### Previous Findings

- ✅ 2 resolved (fixes confirmed in dd2c641, acc3bf5)
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

0 new findings.

### Observation: Deprecation commit dropped

Commit `4652439b` (deprecation frontmatter on 11 cluster `.md` files) is not in the rebased branch. If this was intentional — e.g., deferring deprecation until Phase 3 when skills stop reading the old files — that's a reasonable call. Flagging in case it was an accidental rebase loss.

### Positive Signals

Clean rebase — commit history trimmed from 4 to 3, both fixes preserved exactly.

---
- *Role:* Team-Reviewer

claude/skill-references/github/commands/batch-fetch-reviews.sh
ahoym: I understand that's the design, I want to empirically test it so we can be confident in this solution

ahoym: Good instinct. Here's the quickest empirical test:

**Option A — Direct execution (no Phase 3 needed):**
Open a Claude Code session and run:
```
Read the contents of claude/skill-references/github/commands/batch-fetch-reviews.sh and execute the instructions in it literally against PR #87.
```
Watch whether Claude issues a Write tool call for the jq filter before the `gh api` call. If `tmp/claude-artifacts/jq-filters/jq-filter.jq` gets created, the pattern works.

**Option B — Minimal test skill:**
Create a throwaway `test-inline.md` skill with a single step: `!`cat claude/skill-references/github/commands/batch-fetch-reviews.sh``. Invoke it and observe the tool call sequence.

Option A is faster and doesn't require creating a skill. The key observable is whether the Write tool fires — if it does, `!cat` inlining + `#` directives work as designed. If Claude skips straight to the `gh api` call, the pattern is unreliable and Phase 3 needs a different approach (e.g., numbered steps in the skill body with `!cat` only for the command itself, not the setup instructions).

---
- *Role:* Team-Reviewer

ahoym: Empirical data from this session — I actually hit this exact pattern during cycle 2 addressing:

1. **Write tool fired** — I wrote a jq filter to `tmp/claude-artifacts/jq-filters/jq-filter.jq` following the same Write-first pattern. That part works.
2. **`jq -f` got blocked** — Claude Code's sandbox rejected it with "jq command contains dangerous flags that could execute code or read arbitrary files." The `-cf` variant in this file would hit the same block.

So the `#` comment directives *do* trigger agent behavior when inlined, but the `jq -f`/`-cf` approach has a platform-level blocker. The file's prescribed command (`jq -cf tmp/...`) won't execute in Claude Code sessions as-is.

**Alternatives that worked in this session:**
- Inline jq expressions without `-f` (no string literals to trigger prompts)
- Piping through `jq '...'` with only non-quoted expressions

This suggests Phase 3 may need to restructure these command files — keep the jq filter content inline rather than Write-then-reference, or find a permission-safe way to pass the filter. Worth verifying whether this is a Claude Code version-specific behavior or permanent.

---
- *Persona:* claude-config-author
- *Role:* Addresser

ahoym:
## Review actions

2 implemented (cycle 1) — see inline comments.

| # | File | Comment | Action |
|---|------|---------|--------|
| 3 | batch-fetch-reviews.sh:2 | Verify # comments trigger agent behavior | Clarified — empirical data provided. `jq -f` blocked by sandbox; Phase 3 needs restructuring. Awaiting your decision on approach. |

---
- *Persona:* claude-config-author
- *Role:* Addresser

claude/skill-references/github/commands/batch-fetch-reviews.sh
ahoym: Valuable empirical data — this confirms the design works conceptually but surfaces a real blocker for Phase 3.

**Root cause:** The Write-then-`jq -f` pattern was invented to avoid quoted-string permission prompts (e.g., `"n/a"`, `"(none)"` in jq expressions). But `jq -f` reads an arbitrary file path, which the sandbox treats as a potential code-execution vector — so the fix for one problem created another.

**Recommendation: add a `jq` allow-pattern to `settings.local.json`**

```json
"Bash(jq *)"
```

This is the least-disruptive fix. `jq -f tmp/claude-artifacts/jq-filters/jq-filter.jq` is safe — it reads a known, agent-controlled temp file, not arbitrary paths. The sandbox block is overly conservative for this use case. Adding the allow-pattern lets the existing Write-then-`jq -f` design work without restructuring any `.sh` files.

**Scope:** This affects at least 8 files across both platforms — `batch-fetch-reviews.sh`, `fetch-inline-comments.sh`, `fetch-issue-comments.sh`, `find-approved-reviewers.sh` (GitHub + GitLab each). All use `jq -f` or `jq -cf`/`-rf` with a temp file. Phase 3 prereqs should include verifying this allow-pattern is in place.

**Fallback if allow-pattern isn't viable:** Restructure affected files to inline expressions directly (accepting permission prompts for quoted strings). Larger change but no platform dependency.

---
- *Role:* Team-Reviewer

claude/skill-references/github/commands/batch-fetch-reviews.sh
```diff
@@ -0,0 +1,5 @@
...
+# Then:
+gh api 'repos/{owner}/{repo}/pulls?state=all&sort=created&direction=asc&per_page=<SIZE>&page=<PAGE>' \
+  | jq -cf tmp/claude-artifacts/jq-filters/jq-filter.jq
```
ahoym, line 5: **[claude-config-reviewer]** `jq -cf tmp/claude-artifacts/jq-filters/jq-filter.jq` will be blocked by Claude Code's sandbox in Phase 3. The `-f` flag (reads from file) is flagged as a potential arbitrary code execution vector.

**Affected files (~8 total):** `batch-fetch-reviews.sh`, `fetch-inline-comments.sh`, `fetch-issue-comments.sh`, `find-approved-reviewers.sh` — GitHub and GitLab each.

**Recommended fix:** Add `Bash(jq *)` to `settings.local.json` as a Phase 3 prerequisite. The pattern is safe — the filter file is written by the agent to a known temp path immediately before use.

**Fallback:** Inline jq expressions directly in the command (removes `-f` dependency but requires accepting permission prompts for quoted strings like `"n/a"` and `"(none)"`). Larger change, no platform dependency.

---
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: feat: extract .sh platform command files (#83 — phase 1)

New finding surfaced by empirical testing in the addressing session. No new commits since last pass (`acc3bf5`).

Reviewers (this cycle): claude-config-reviewer
Carried forward: none

### Previous Findings

- ✅ 2 resolved (dd2c641, acc3bf5)
- 👍 0 acknowledged
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

1 new finding — see inline comment on `batch-fetch-reviews.sh`.

### ⚠️ Phase 3 Prerequisite: `jq -f` sandbox block

Empirical testing confirmed `#` comment directives DO trigger Write tool calls when inlined via `!cat`. However, `jq -f tmp/...` (and `-cf`/`-rf` variants) is blocked by Claude Code's sandbox as a potential code-execution vector. This affects ~8 files across both platforms that use the Write-then-`jq -f` pattern.

Recommended fix: add `Bash(jq *)` to `settings.local.json` before Phase 3. The allow-pattern is safe here — the temp file path is agent-controlled. See inline comment for alternatives if that's not viable.

---
- *Role:* Team-Reviewer

claude/skill-references/github/commands/batch-fetch-reviews.sh
ahoym: Agreed — `Bash(jq *)` is the right fix. The filter file is agent-controlled and written to a known temp path immediately before use, so the sandbox block is overly conservative here.

This is a Phase 3 prerequisite, not a Phase 1 change — the `.sh` command files in this PR are correct as designed, they just need the allow-pattern in place before Phase 3 wires them up. I'll note it as a prerequisite in the Phase 3 tracking.

---
- *Persona:* claude-config-author
- *Role:* Addresser

claude/skill-references/github/commands/batch-fetch-reviews.sh
ahoym: Covered in the thread above — agreed on `Bash(jq *)` as Phase 3 prerequisite. The 8-file scope assessment is useful for Phase 3 planning.

---
- *Persona:* claude-config-author
- *Role:* Addresser

claude/skill-references/gitlab/commands/post-review-comments.sh
```diff
@@ -0,0 +1,29 @@
...
+# Step 1: Get SHAs and MR internal ID:
+# glab api projects/:id/merge_requests/<number>/versions | jq '.[0] | {base_commit_sha, head_commit_sha}'
+# glab mr view <number> --output json | jq '.id'
+# The MR id (not iid) is needed for GraphQL noteableId: "gid://gitlab/MergeRequest/<id>"
+
+# Step 2: Post each inline comment via GraphQL:
+# For complex mutations, write query to .graphql file and use uppercase -F:
+#   glab api graphql -F query=@tmp/claude-artifacts/change-request-replies/gql-1.graphql
+# Lowercase -f query=@file does NOT read the file — passes literal "@file" string.
+glab api graphql -f query='mutation {
```
ahoym, line 13: comment says to use -F but this command is literally lowercase -f

ahoym: The command is correct — the apparent contradiction is a comment structure issue.

The file documents two separate approaches:

1. **File-based** (lines 10-11): Use `-F query=@file` (uppercase) to read the mutation from a `.graphql` file. Recommended for complex mutations.

2. **Inline** (line 13+): Use `-f query='mutation {...}'` (lowercase) with the literal mutation string directly in the command. This is what the command below demonstrates.

Line 12 is a **gotcha warning for a specific mistake**: don't try `glab api graphql -f query=@file` (lowercase + @file). That combination won't read the file — it passes the literal string `"@file"`. You'd need uppercase `-F` for that.

So `-f` with an inline literal (line 13) ≠ `-f` with `@file` (the mistake line 12 warns against). They look the same flag but do different things based on the value.

That said, the comment structure *does* read as contradictory — the transition from "use uppercase -F" to immediately showing lowercase `-f` is jarring without explicit framing of "alternatively, inline it." A comment like `# Alternatively, inline the mutation directly:` before line 13 would disambiguate. Worth noting for Phase 3 when skills start reading this file.

---
- *Role:* Team-Reviewer

ahoym: Valid catch — the transition from "use uppercase `-F` for file-based" to immediately showing lowercase `-f` with an inline literal is confusing without framing.

Adding a comment before the inline example to disambiguate: `# Alternatively, inline the mutation directly (lowercase -f with literal string):`. This makes it clear we're showing a second approach, not contradicting the first.

---
- *Persona:* claude-config-author
- *Role:* Addresser

ahoym: Fixed in 31c3244

---
- *Persona:* claude-config-author
- *Role:* Addresser

ahoym (commented):
## Team Re-review: feat: extract .sh platform command files (#83 — phase 1)

One new commit (`31c3244`) adding a transitional comment to `gitlab/commands/post-review-comments.sh`. Reviewed — no new findings.

Reviewers (this cycle): claude-config-reviewer
Carried forward: none

### Previous Findings

- ✅ 3 resolved (dd2c641, acc3bf5, 31c3244)
- 👍 1 acknowledged (jq -f sandbox block — Phase 3 prereq, no Phase 1 fix)
- 🔄 0 partially addressed
- ❌ 0 not addressed

### New Findings

0 new findings.

### Positive Signals

The `# Alternatively, inline the mutation directly (lowercase -f with literal string):` addition lands exactly right — makes the two-option structure explicit without restructuring the file. The operator Q&A loop surfaced a real documentation gap before Phase 3 wires this up.

---
- *Role:* Team-Reviewer

PR #89, opened 2026-04-13
feat: add platform symlink to setup-claude.sh (#83 — phase 4)

ahoym:
## Summary

- `setup-claude.sh` gains platform detection: prompts for `github`/`gitlab` on first run, caches in `~/.claude/.platform`
- Creates symlink: `~/.claude/platform-commands` -> `skill-references/{github,gitlab}/commands/`
- Skills use `!`cat ~/.claude/platform-commands/<cmd>.sh`` to inline the correct platform's commands at preprocessing time
- Idempotent: skips if symlink already correct, backs up if different

## Test plan

- [ ] Run `setup-claude.sh` — verify platform prompt appears and answer `github`
- [ ] Re-run — verify it reads from `.platform` cache (no prompt)
- [ ] `cat ~/.claude/platform-commands/create-review.sh` — verify symlink resolves

Relates to https://github.com/ahoym/dotfiles/issues/83

### DIFF
--- a/setup-claude.sh
+++ b/setup-claude.sh
@@ -40,3 +40,42 @@ for item in "${ITEMS[@]}"; do
   ln -s "$src" "$dest"
   echo "  link   $item -> $src"
 done
+
+# --- Platform-specific commands symlink ---
+# Creates ~/.claude/platform-commands -> skill-references/{github,gitlab}/commands/
+# Skills use !`cat ~/.claude/platform-commands/<cmd>.sh` to inline platform commands.
+
+PLATFORM_FILE="$TARGET_DIR/.platform"
+PLATFORM_LINK="$TARGET_DIR/platform-commands"
+
+if [ -f "$PLATFORM_FILE" ]; then
+  PLATFORM=$(cat "$PLATFORM_FILE")
+  echo "  platform: $PLATFORM (cached in .platform)"
+else
+  printf "  Which platform does this machine use? [github/gitlab]: "
+  read -r PLATFORM
+  case "$PLATFORM" in
+    github|gitlab) ;;
+    *) echo "  error: expected github or gitlab, got $PLATFORM"; exit 1 ;;
+  esac
+  echo "$PLATFORM" > "$PLATFORM_FILE"
+  echo "  platform: $PLATFORM (saved to .platform)"
+fi
+
+PLATFORM_SRC="$DOTFILES_CLAUDE_DIR/skill-references/$PLATFORM/commands"
+
+if [ ! -d "$PLATFORM_SRC" ]; then
+  echo "  skip   platform-commands ($PLATFORM_SRC not found)"
+else
+  if [ -L "$PLATFORM_LINK" ] && [ "$(readlink "$PLATFORM_LINK")" = "$PLATFORM_SRC" ]; then
+    echo "  ok     platform-commands -> $PLATFORM/commands"
+  else
+    if [ -e "$PLATFORM_LINK" ] || [ -L "$PLATFORM_LINK" ]; then
+      backup="$PLATFORM_LINK.backup.$(date +%s)"
+      echo "  backup platform-commands -> $(basename $backup)"
+      mv "$PLATFORM_LINK" "$backup"
+    fi
+    ln -s "$PLATFORM_SRC" "$PLATFORM_LINK"
+    echo "  link   platform-commands -> $PLATFORM/commands"
+  fi
+fi
--- a/setup-claude.sh
+++ b/setup-claude.sh
@@ -50,6 +50,10 @@ PLATFORM_LINK="$TARGET_DIR/platform-commands"
 
 if [ -f "$PLATFORM_FILE" ]; then
   PLATFORM=$(cat "$PLATFORM_FILE")
+  case "$PLATFORM" in
+    github|gitlab) ;;
+    *) echo "  error: invalid cached platform '$PLATFORM'; delete .platform and re-run"; exit 1 ;;
+  esac
   echo "  platform: $PLATFORM (cached in .platform)"
 else
   printf "  Which platform does this machine use? [github/gitlab]: "
@@ -72,7 +76,7 @@ else
   else
     if [ -e "$PLATFORM_LINK" ] || [ -L "$PLATFORM_LINK" ]; then
       backup="$PLATFORM_LINK.backup.$(date +%s)"
-      echo "  backup platform-commands -> $(basename $backup)"
+      echo "  backup platform-commands -> $(basename "$backup")"
       mv "$PLATFORM_LINK" "$backup"
     fi
     ln -s "$PLATFORM_SRC" "$PLATFORM_LINK"


setup-claude.sh
```diff
@@ -40,3 +40,42 @@ for item in "${ITEMS[@]}"; do
...
+
+# --- Platform-specific commands symlink ---
+# Creates ~/.claude/platform-commands -> skill-references/{github,gitlab}/commands/
+# Skills use !`cat ~/.claude/platform-commands/<cmd>.sh` to inline platform commands.
+
+PLATFORM_FILE="$TARGET_DIR/.platform"
+PLATFORM_LINK="$TARGET_DIR/platform-commands"
+
+if [ -f "$PLATFORM_FILE" ]; then
+  PLATFORM=$(cat "$PLATFORM_FILE")
```
ahoym, line 52: **[correctness-reviewer, security-reviewer] Medium — cached `.platform` value not re-validated**

The cached branch reads and uses `$PLATFORM` directly without passing through the `case github|gitlab` whitelist:

```bash
PLATFORM=$(cat "$PLATFORM_FILE")
echo "  platform: $PLATFORM (cached in .platform)"  # no validation
```

If the file was manually edited, contains a typo (e.g. `gittub`), or has CRLF line endings (`github\r` from a Windows editor), `PLATFORM_SRC` becomes a nonexistent path. The `[ ! -d ]` guard then silently emits "skip platform-commands" with no indication that the cached value is the problem.

Suggestion — apply the same validation after reading from cache:

```bash
PLATFORM=$(cat "$PLATFORM_FILE")
case "$PLATFORM" in
  github|gitlab) ;;
  *) echo "  error: invalid cached platform '$PLATFORM'; delete .platform and re-run"; exit 1 ;;
esac
```

---
- *Persona:* correctness-reviewer, security-reviewer
- *Role:* Team-Reviewer

ahoym (commented):
## Team Review: feat: add platform symlink to setup-claude.sh

The platform symlink feature is logically sound — idempotency checks, dangling-symlink handling, and epoch-timestamped backups all reflect careful defensive design. Two themes emerged across both reviews: the cached `.platform` path bypasses the only input validator (creating a silent trust gap), and the absence of exit-code guards on `mv` and the `.platform` write allows partial failures to appear as success.

Reviewed by: correctness-reviewer, security-reviewer

### Findings

- **[correctness-reviewer, security-reviewer]** Cached `.platform` read bypasses the only input whitelist — silent skip on corrupt/edited file
- **[correctness-reviewer]** `mv` before `ln -s` has no exit-code check — silent partial failure if move fails
- **[security-reviewer]** Whitelist lives only in the `case` block, not at the path construction site
- **[correctness-reviewer, security-reviewer]** Unquoted `$backup` in `$(basename $backup)` — latent word-split risk
- **[correctness-reviewer]** No `set -euo pipefail` — `mv`, `cat`, and `.platform` write failures all continue silently (summary-only)
- **[correctness-reviewer]** `.platform` write not checked — false "saved" message on disk-full or permission error (summary-only)
- **[security-reviewer]** `read -r PLATFORM` blocks indefinitely when stdin is not a terminal (summary-only)

### Positive Signals

- TOCTOU avoidance on the happy path: `readlink` equality check before replacing prevents unnecessary churn on idempotent re-runs
- Dangling-symlink guard (`-e || -L`) correctly handles the edge case where `-e` is false but `-L` is true — subtle but right
- Epoch timestamp on backups prevents clobbering previous backup files — good defensive design
- Whitelist-only `case` statement on interactive input fails fast with a clear error message

---
- *Persona:* correctness-reviewer, security-reviewer
- *Role:* Team-Reviewer

setup-claude.sh
```diff
@@ -40,3 +40,42 @@ for item in "${ITEMS[@]}"; do
...
+  read -r PLATFORM
+  case "$PLATFORM" in
+    github|gitlab) ;;
+    *) echo "  error: expected github or gitlab, got $PLATFORM"; exit 1 ;;
+  esac
+  echo "$PLATFORM" > "$PLATFORM_FILE"
+  echo "  platform: $PLATFORM (saved to .platform)"
+fi
+
+PLATFORM_SRC="$DOTFILES_CLAUDE_DIR/skill-references/$PLATFORM/commands"
```
ahoym, line 69: **[security-reviewer] Medium — whitelist only at input, not at path construction site**

`PLATFORM_SRC` is constructed directly from `$PLATFORM` with no assertion at the point of use. The `case` block is the sole guard, and it lives in a separate code branch. A future refactor that loosens or replaces the `case` would silently remove the only protection against path traversal.

Consider adding a local guard immediately before path construction:

```bash
[[ "$PLATFORM" == 'github' || "$PLATFORM" == 'gitlab' ]] || { echo "  error: invalid platform '$PLATFORM'"; exit 1; }
PLATFORM_SRC="$DOTFILES_CLAUDE_DIR/skill-references/$PLATFORM/commands"
```

This makes the invariant local to the site where it matters and survives future changes to the input-gathering block.

---
- *Persona:* correctness-reviewer, security-reviewer
- *Role:* Team-Reviewer

setup-claude.sh
```diff
@@ -40,3 +40,42 @@ for item in "${ITEMS[@]}"; do
...
+if [ ! -d "$PLATFORM_SRC" ]; then
+  echo "  skip   platform-commands ($PLATFORM_SRC not found)"
+else
+  if [ -L "$PLATFORM_LINK" ] && [ "$(readlink "$PLATFORM_LINK")" = "$PLATFORM_SRC" ]; then
+    echo "  ok     platform-commands -> $PLATFORM/commands"
+  else
+    if [ -e "$PLATFORM_LINK" ] || [ -L "$PLATFORM_LINK" ]; then
+      backup="$PLATFORM_LINK.backup.$(date +%s)"
+      echo "  backup platform-commands -> $(basename $backup)"
+      mv "$PLATFORM_LINK" "$backup"
```
ahoym, line 80: **[correctness-reviewer] Medium — unchecked `mv` before `ln -s`**

```bash
mv "$PLATFORM_LINK" "$backup"
ln -s "$PLATFORM_SRC" "$PLATFORM_LINK"
```

If `mv` fails (permissions, cross-device move, read-only mount), the script doesn't abort. `ln -s` then also fails because `$PLATFORM_LINK` still exists. The `mv` failure message is the only signal — easy to miss in a setup script's output stream, and the user may believe setup completed successfully.

Add an explicit abort:

```bash
mv "$PLATFORM_LINK" "$backup" || { echo "  error: failed to back up existing link"; exit 1; }
```

Alternatively, `set -euo pipefail` at the top of the script catches this (and the `.platform` write redirect failure at line 61) automatically.

---
- *Persona:* correctness-reviewer, security-reviewer
- *Role:* Team-Reviewer

setup-claude.sh
```diff
@@ -40,3 +40,42 @@ for item in "${ITEMS[@]}"; do
...
+
+if [ ! -d "$PLATFORM_SRC" ]; then
+  echo "  skip   platform-commands ($PLATFORM_SRC not found)"
+else
+  if [ -L "$PLATFORM_LINK" ] && [ "$(readlink "$PLATFORM_LINK")" = "$PLATFORM_SRC" ]; then
+    echo "  ok     platform-commands -> $PLATFORM/commands"
+  else
+    if [ -e "$PLATFORM_LINK" ] || [ -L "$PLATFORM_LINK" ]; then
+      backup="$PLATFORM_LINK.backup.$(date +%s)"
+      echo "  backup platform-commands -> $(basename $backup)"
```
ahoym: **[correctness-reviewer, security-reviewer] Low — unquoted `$backup` in `$(basename $backup)`**

`$backup` is unquoted inside the command substitution. Bash performs word splitting and glob expansion before the argument reaches `basename`. While `$backup` is derived from controlled values (a path + epoch timestamp) and won't produce spaces or globs today, the pattern is a latent fragility:

```bash
echo "  backup platform-commands -> $(basename "$backup")"
```

---
- *Persona:* correctness-reviewer, security-reviewer
- *Role:* Team-Reviewer

setup-claude.sh
ahoym: Good catch. The cached read does bypass the whitelist, and a corrupted `.platform` file (or CRLF line endings) would silently produce a bad path. Implementing the suggested validation after the cache read.

---
- *Persona:* claude-config-author
- *Role:* Addresser

setup-claude.sh
ahoym: With comment #1's fix implemented, both branches (cached + interactive) now validate `$PLATFORM` through the `case` whitelist before reaching line 65. That means `$PLATFORM` is guaranteed valid at the path construction site without a third guard.

Adding another check here would be belt-and-suspenders-and-a-third-belt. The invariant is already established by the time control reaches `PLATFORM_SRC=`. If a future refactor removes validation from both branches, a local guard here wouldn't save it either — the `.platform` write on line 61 would already have persisted the bad value.

Also noting the suggested `[[ ]]` syntax is inconsistent with the script's `[ ]` POSIX-style usage throughout.

---
- *Persona:* claude-config-author
- *Role:* Addresser

setup-claude.sh
ahoym: The script already has `set -e` on line 6. Under `set -e`, commands in `then`/`else` blocks (as opposed to `if` conditions) do exit on failure — so `mv` failing here already aborts the script.

The `set -euo pipefail` suggestion would add `-u` (unbound variable) and `-o pipefail` on top, which are good hardening but orthogonal to the `mv` concern specifically. No change needed for this particular issue.

---
- *Persona:* claude-config-author
- *Role:* Addresser

setup-claude.sh
ahoym: Good eye — the loop section on line 36 already quotes `"$backup"` properly. This is just an inconsistency in the new code. Fixing.

---
- *Persona:* claude-config-author
- *Role:* Addresser

ahoym:
## Review actions

2 implemented — see inline comments.

| # | File | Suggestion | Action |
|---|------|------------|--------|
| 2 | setup-claude.sh:65 | Add guard at path construction site | Awaiting your decision (disagree) — comment #1's fix covers both branches |
| 3 | setup-claude.sh:76 | Add exit-code check on `mv` | Awaiting your decision (disagree) — `set -e` on line 6 already handles this |

---
- *Persona:* claude-config-author
- *Role:* Addresser

setup-claude.sh
ahoym: Fixed in 1a4e45a

---
- *Persona:* claude-config-author
- *Role:* Addresser

setup-claude.sh
ahoym: Fixed in 1a4e45a

---
- *Persona:* claude-config-author
- *Role:* Addresser

setup-claude.sh
ahoym: Agreed — the cache fix makes this redundant. With validation now in both branches (cached and interactive), `$PLATFORM` is guaranteed to be `github` or `gitlab` before control ever reaches line 65. The original concern was a single guard in one code path, which is no longer the case.

Closing this finding. ✅

---
- *Persona:* correctness-reviewer, security-reviewer
- *Role:* Team-Reviewer

setup-claude.sh
ahoym: You're right — the review missed `set -e` on line 6. Under `set -e`, a failing `mv` inside an `else` block will abort the script, which is exactly the protection this finding was after. The explicit `|| { exit 1; }` is redundant given the existing guard.

The `-uo pipefail` additions would add unbound-variable detection and pipeline failure propagation, but those are orthogonal to the `mv` concern and not required here. No change needed. ✅

---
- *Persona:* correctness-reviewer, security-reviewer
- *Role:* Team-Reviewer

ahoym (commented):
## Team Re-review: feat: add platform symlink to setup-claude.sh

Commit `1a4e45a` resolves the two agreed findings — the cached `.platform` validation and the unquoted `$backup`. The two disagreements are valid pushbacks: the cache fix closes the defense-in-depth concern, and `set -e` (line 6) already handles the `mv` exit-code case the review missed.

Reviewers (this cycle): orchestrator direct (targeted fix, no new domain coverage needed)
Carried forward: correctness-reviewer, security-reviewer

### Previous Findings

- ✅ 2 resolved (code change verified)
- 🔄 2 closed via valid operator pushback (findings superseded)

### New Findings

0 new findings in `1a4e45a` — the changes are well-scoped and correct.

### Positive Signals

- Cache validation now uses the same `case` pattern as the interactive branch, including the tailored error message pointing to `.platform` — a clean, consistent fix
- The `set -e` coverage was already in place; good to have this surfaced explicitly for future readers

---
- *Persona:* correctness-reviewer, security-reviewer
- *Role:* Team-Reviewer

PR #109, opened 2026-05-17
Add learnings-suggest hook: federated keyword matching & section-level indexing

ahoym:
## Summary

Implements a comprehensive learnings suggestion system that matches user prompts against a federated corpus of learnings, guidelines, and skill references. The system uses dual-layer indexing (section-level preferred, file-level fallback) to inject contextual `<learnings-suggestions>` hints into prompts, along with telemetry and staleness detection.

## Key Changes

- **UserPromptSubmit hook** (`learnings-suggest`): Aho-Corasick-based keyword matching against both section-level (from `sections.json`) and file-level (from `.keyword-index.json`) indexes. Scores hits by keyword weight, filters by staleness, and respects quoted-term bypass. Injects up to 3 suggestions as a hint block. Never blocks a prompt.

- **Section-level indexing** (`learnings-index-build`): Walks the federated learnings corpus, parses Markdown with pulldown-cmark, extracts H2/H3 sections with line ranges, and builds an inverted keyword index. Supports explicit file-level keywords (via `**Keywords:**` convention) and deduplicates section anchors. Outputs `~/.claude/claude-artifacts/ast/sections.json`.

- **Read telemetry** (`learnings-read-log`): PostToolUse hook that logs all Read operations on learnings/guidelines/skill-references files to `reads.jsonl` for offline hit-rate analysis.

- **Staleness detection** (`learnings-staleness.py`): SessionStart hook that warns when the keyword index is stale (≥5 files changed since last rebuild), with git-based anchor tracking.

- **Analysis tooling** (`analyze.py`): Cross-references suggest.jsonl and reads.jsonl to compute tier hit rates, identify top performers, and surface noise/coverage gaps.

- **Build system**: Rust Cargo project with platform-specific binaries (aarch64/x86_64, Darwin/Linux), idempotent bootstrap script, and optional cross-compilation via cargo-zigbuild.

- **Hook registration**: Added to `settings.json` under `hooks.UserPromptSubmit`, `hooks.PostToolUse`, and `hooks.SessionStart`.

- **Setup integration**: Updated `setup-claude.sh` to symlink the `hooks/` directory and run the bootstrap build.

## Notable Implementation Details

- **Dual indexing strategy**: Section-level hits are preferred (more precise), but file-level hits supplement them and serve dense files where section extraction may be incomplete. Covered paths are deduplicated.

- **Staleness downgrade**: If a source file is newer than `sections.json`, section references are downgraded to file-level (anchor/lines dropped) to avoid stale line numbers.

- **Quoted-term bypass**: Keywords in double quotes force inclusion even below the MIN_SCORE threshold, enabling explicit learnings requests.

- **Telemetry design**: Prompt hashes (not full text) are logged for privacy; session IDs enable cross-referencing with Read logs; tier classification (strong ≥3, weak <3) supports hit-rate stratification.

- **Silent failure**: All hooks are designed to never block a prompt/session/tool call — any I/O or parsing error results in a no-op.

- **Provider resolution**: Supports both `learnings-providers.json` entries (localPath, projectLocal) and runtime discovery via git anchors for commit-based staleness tracking.

https://claude.ai/code/session_01LnbK3YBtPx3GQUbCrkc43S


claude/hooks/learnings-suggest/src/bin/index-build.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+#[derive(Clone)]
+struct Provider {
+    name: String,
+    base: PathBuf,  // The directory that contains the learnings tree
+    root: PathBuf,  // The path that rel-paths in indexes are relative to (parent of base)
+}
+
+fn providers() -> Vec<Provider> {
```
ahoym: **[HIGH] [architecture-reviewer] `providers()` duplicated with a hidden behavioral gap**

`providers()` (and `home()`, `expand_tilde()`, `Provider`) are defined here verbatim as they are in `src/main.rs`, but with one key behavioral difference: this version omits the `projectLocal` branch (lines 73-82 of `main.rs`). The result is that project-local learnings are never section-indexed even though `run()` in `main.rs` will match against them via the file-level fallback. Any engineer adding a third binary will copy-paste a third version and the asymmetry will silently widen.

Suggested path: move `Provider`, `home()`, `expand_tilde()`, and the core provider loop into `src/lib.rs`. The `projectLocal` branch can be opt-in via a boolean parameter (`providers(include_project_local: bool)`), making the behavioral split explicit and visible.

---
- *Persona:* architecture-reviewer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/src/bin/index-build.rs
```diff
@@ -0,0 +1,430 @@
...
+use pulldown_cmark::{Event, HeadingLevel, Parser, Tag, TagEnd};
+use serde_json::{json, Value};
+use std::collections::{BTreeMap, HashMap, HashSet};
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+const MIN_SECTION_LINES: usize = 8;
+const MAX_KEYWORDS_PER_SECTION: usize = 12;
+const SCHEMA_VERSION: u32 = 1;
```
ahoym: **[MEDIUM] [architecture-reviewer] SCHEMA_VERSION type mismatch: `u32` here, `u64` in `main.rs`**

`index-build.rs` writes `SCHEMA_VERSION` as a `u32`; `main.rs` reads it back via `Value::as_u64()` and compares against its own `u64` constant. At value 1 both serialize identically, but this is a latent trap: two constants with the same name and different types in files that must stay in sync. If the reader switches to `as_u32()` for consistency or the version is bumped in one file and not the other, the schema check silently fails.

Prefer `u64` to match the reader. This is resolved naturally if both files are unified via `src/lib.rs`.

---
- *Persona:* architecture-reviewer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/src/bin/index-build.rs
```diff
@@ -0,0 +1,430 @@
...
+    "what", "when", "where", "why", "how", "which", "who",
+    "one", "two", "three", "each", "every", "all", "some", "any", "none",
+    "more", "most", "less", "very", "much", "many", "just", "only",
+    "also", "still", "always", "never", "often", "usually",
+    "use", "used", "using", "uses", "make", "makes", "made", "making",
+    "see", "now", "get", "got", "set", "run", "ran", "via", "such",
+    "e.g", "i.e",
+];
+
+fn home() -> PathBuf {
```
ahoym: **[MEDIUM] [architecture-reviewer] `home()` and `expand_tilde()` duplicated verbatim from `main.rs`**

Byte-for-byte copies. The multi-binary `Cargo.toml` already exists; adding a `[lib]` target costs one line and makes both utilities importable from `lib.rs`. Bug fixes to tilde expansion (e.g., handling `$HOME` edge cases) currently require patching two files.

---
- *Persona:* architecture-reviewer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/src/bin/index-build.rs
```diff
@@ -0,0 +1,430 @@
...
+            total_bytes,
+            header,
+            &body_buf,
+        );
+        let _ = total_lines;
+    }
+    sections
+}
+
+fn anchor_commit() -> Option<String> {
```
ahoym: **[LOW] [architecture-reviewer] `anchor_commit()` re-parses `learnings-providers.json` unnecessarily**

`main()` already has a fully-loaded `provs: Vec<Provider>` slice. `anchor_commit()` calls `providers()` again, causing a second read and parse of `learnings-providers.json`. Pass the existing slice in: `fn anchor_commit(provs: &[Provider]) -> Option<String>`.

---
- *Persona:* architecture-reviewer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/src/main.rs
```diff
@@ -0,0 +1,530 @@
...
+
+fn build_matcher(keys: &[&str]) -> Option<aho_corasick::AhoCorasick> {
+    AhoCorasickBuilder::new()
+        .match_kind(MatchKind::LeftmostLongest)
+        .ascii_case_insensitive(true)
+        .build(keys)
+        .ok()
+}
+
+fn run() -> Option<()> {
```
ahoym: **[MEDIUM] [architecture-reviewer] `run()` is ~200 lines covering six inlined phases**

The function mixes payload parsing, provider loading, section-level matching, file-level matching, deduplication, quoted-term bypass, telemetry, and output formatting — all in one body. The phase comments are helpful but don't substitute for extraction: each phase cannot be tested in isolation as-is.

Suggested extraction: `match_sections()`, `match_files()`, `merge_and_dedup()`, and `format_block()` — reducing `run()` to a readable 30-line coordinator with each phase independently testable.

---
- *Persona:* architecture-reviewer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/src/main.rs
```diff
@@ -0,0 +1,530 @@
...
+        }
+    }
+    let mut v: FileIndexEntries = merged.into_iter().collect();
+    v.sort_by(|a, b| a.0.cmp(&b.0));
+    v
+}
+
+// ---------- Prompt helpers ----------
+
+fn extract_quoted(prompt: &str) -> HashSet<String> {
```
ahoym, line 238: **[LOW] [architecture-reviewer] `extract_quoted()` matches ASCII double-quotes only**

On macOS with autocorrect, typed quotes may become `\u201c`/`\u201d` (curly quotes), which silently bypass the bypass. The documented convention (quoted terms force inclusion below MIN_SCORE) then works differently across input methods. At minimum, a comment noting the ASCII-only constraint would prevent confusion: `// Only matches ASCII ".."; curly quotes from autocorrect are not captured`.

---
- *Persona:* architecture-reviewer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/bootstrap.sh
```diff
@@ -0,0 +1,111 @@
...
+  if [ "$OS" = "Darwin" ] && command -v brew >/dev/null 2>&1; then
+    brew install rust
+  else
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
+      | sh -s -- -y --default-toolchain stable --profile minimal
+    export PATH="$HOME/.cargo/bin:$PATH"
+  fi
+}
+
+stage_binaries() {
```
ahoym, line 61: **[LOW] [architecture-reviewer, claude-config-reviewer] `any_stale()` doesn't check `Cargo.lock`**

Dependency version bumps (patch releases of `aho-corasick`, `pulldown-cmark`) land only in `Cargo.lock`, not `Cargo.toml`. The freshness check will incorrectly report the binary as up-to-date after a lockfile-only update.

Add: `[ "$bin" -nt "$DIR/Cargo.lock" ] || return 0` alongside the existing `Cargo.toml` check.

Also: the return-value convention (0 = stale, `|| return 0` short-circuits) is hard to audit on first read. A one-line comment helps: `# Returns 0 (true in bash) when stale; exits early at each stale condition.`

---
- *Persona:* architecture-reviewer, claude-config-reviewer
- *Role:* Team-Reviewer

ahoym (commented):
## Team Review: Add learnings-suggest hook: federated keyword matching & section-level indexing

A well-designed learnings suggestion system with solid fail-safe discipline throughout — no hook ever blocks a prompt, the atomic index write is correct, and the two-tier matching architecture (section-level preferred, file-level supplement) is clean and extensible. The main areas to address are a behavioral gap from duplicated provider loading logic, a SCHEMA_VERSION type mismatch that's latent now but a trap on the next bump, and the global scope of hook registration which adds per-prompt overhead in every project session.

Reviewed by: claude-config-reviewer, architecture-reviewer, python-engineer

### Findings

14 finding(s) — see inline comments.

### Positive Signals

- Fail-safe discipline is excellent: every wrapper shell script exits 0 on a missing or non-executable binary; the Python hooks catch all exceptions at the outer boundary with a clear `# noqa: BLE001` rationale. The system never blocks a session under any failure mode.
- Atomic index write (tmp file + rename) in `index-build.rs` correctly prevents a partial `sections.json` from poisoning the hook on concurrent reads.
- The staleness downgrade in `run()` — demoting a section hit to file-level when the source file is newer than `sections.json` — is a thoughtful degradation path that keeps suggestions usable without requiring a full rebuild after every content edit.
- `setup-claude.sh` is updated to include `hooks/` in the symlink list and runs `bootstrap.sh` idempotently, preserving the single-step provisioning invariant.
- `FILTER_PREFIXES` is a single named constant; adding a new exclusion is a one-line change in an obvious location.
- The `load_jsonl` function in `analyze.py` handles missing files, empty lines, and per-line `JSONDecodeError` gracefully — correct for a log-file reader where individual record corruption should never abort the whole analysis.

---
- *Persona:* claude-config-reviewer, architecture-reviewer, python-engineer
- *Role:* Team-Reviewer

setup-claude.sh
```diff
@@ -83,3 +83,24 @@ else
     echo "  link   platform-commands -> $PLATFORM/commands"
   fi
 fi
+
+# --- Build the learnings-suggest hook binary for this platform ---
```
ahoym, line 87: **[LOW] [architecture-reviewer] `setup-claude.sh` duplicates the platform case statement from `rebuild-index.sh`**

The platform dispatch to select the index builder binary now lives in three places: `learnings-suggest.sh`, `rebuild-index.sh`, and here. `rebuild-index.sh` already handles platform selection and the silent no-op pattern. This block could simply call `rebuild-index.sh` after bootstrap completes:

```bash
"$HOOK_BOOTSTRAP" || echo "  warn   learnings-suggest build failed — hook will no-op"
bash "$DOTFILES_CLAUDE_DIR/hooks/learnings-suggest/rebuild-index.sh" 2>&1 | sed 's/^/  index  /' || true
```

---
- *Persona:* architecture-reviewer
- *Role:* Team-Reviewer

claude/settings.json
```diff
@@ -136,5 +136,29 @@
     "type": "command",
     "command": "bash ~/.claude/statusline-command.sh"
   },
+  "hooks": {
```
ahoym, line 139: **[MEDIUM] [claude-config-reviewer] Hooks registered globally fire on every project session**

Hooks in `settings.json` fire in every Claude Code session regardless of CWD. The `UserPromptSubmit` hook spawns a binary on every prompt in every project — including Java, frontend, or other codebases that have no learnings corpus. This adds per-prompt process spawn overhead (~1-2ms) and may surface irrelevant suggestions in non-dotfiles contexts.

From `hooks.md`: *"Scope hooks to contexts where they're needed — inject into worktree-level `settings.local.json` instead of user-level settings."*

Two options:
1. Move hook registrations to `settings.local.json` (machine-local, gitignored) and have `setup-claude.sh` inject them there.
2. Add an early-exit guard in `learnings-suggest.sh`: exit 0 when `~/.claude/learnings-providers.json` is absent, limiting blast radius to machines that have run `setup-claude.sh`.

---
- *Role:* Team-Reviewer

claude/guidelines/context-aware-learnings.md
```diff
@@ -16,12 +16,44 @@ All gates are mandatory when their trigger fires. No exceptions.
...
 | **Implementation start** | Before executing approved plan | Glob persona dirs → tech stack | Yes — activate match |
 | **Review entry** | Invoking a review/audit skill that doesn't self-load learnings (e.g., native `/simplify`, `/security-review`) BEFORE launching review subagents. Skip if the skill's instructions already include a "load learnings" step (`git:code-review-request`, `git:team-review-request`, `git:address-request-comments` all handle this internally) | Glob filenames + grep content → quality, testing, security, perf, language-specific gotchas relevant to the diff | No |
-| **Keyword** | Domain keyword or quoted term in user message | Glob filenames → unloaded matches. Quoted terms bypass dedup. | No |
-| **Domain shift** | User message introduces new domain | Glob filenames → new domain keywords | No |
-| **Pre-edit check** | First Edit/Write in unloaded tech domain | Glob filenames → file's tech domain | No |
+| **Keyword** *(hook-owned)* | Domain keyword or quoted term in user message | Handled by `learnings-suggest` `UserPromptSubmit` hook — see below | No |
```
ahoym: **[MEDIUM] [claude-config-reviewer] Hook-owned gates don't document agent fallback when binary is absent**

The updated gate table says `Handled by learnings-suggest UserPromptSubmit hook — see below` for keyword and domain-shift gates. But if the binary is absent (unsupported platform, failed bootstrap), those gates simply don't fire — the agent won't know to fall back to the manual pipeline search. The `Pre-edit check` row correctly notes `Today: agent-owned. Iteration 2: PreToolUse hook` — the keyword/domain-shift rows need equivalent fallback language.

Suggested addition: `Handled by learnings-suggest hook when present; agent falls back to manual pipeline search when binary is absent.`

---
- *Role:* Team-Reviewer

claude/guidelines/context-aware-learnings.md
```diff
@@ -16,12 +16,44 @@ All gates are mandatory when their trigger fires. No exceptions.
...
+| **Domain shift** *(hook-owned)* | User message introduces new domain | Same hook | No |
+| **Pre-edit check** *(hook-owned, iteration 2)* | First Edit/Write in unloaded tech domain | Today: agent-owned. Iteration 2: `PreToolUse` hook on Edit/Write | No |
 
 **Dedup**: don't re-load files already read this session. When context compression makes history uncertain, err toward re-checking.
 
+## Hook-suggested learnings
```
ahoym, line 28: **[MEDIUM] [claude-config-reviewer] Hook-suggested learnings section could be compressed**

This section is 40+ lines of always-on context. The Tier interpretation bullets (strong/weak/ignore/quoted) are a natural 2-column table. The offset+limit mechanics in the Path forms section are already documented in CLAUDE.md's Read Tool section — loading both is token duplication.

Suggested trim (targeting ~28 lines from ~42):
- Collapse Tier interpretation into `| Tier | Action |` table
- Cut the `Read with offset=start, limit=(end-start+1)` explanation from Path forms (already in CLAUDE.md)
- Keep the Staleness paragraph as-is

Guidelines are always-on context; every token costs on every session.

---
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/analyze.py
```diff
@@ -0,0 +1,161 @@
...
+    args = ap.parse_args()
+
+    suggests = load_jsonl(SUGGEST_LOG)
+    reads = load_jsonl(READS_LOG)
+
+    if not suggests:
+        print(f"No suggestion log at {SUGGEST_LOG}", file=sys.stderr)
+        sys.exit(1)
+
+    cutoff = max(s.get("ts", 0) for s in suggests) - args.days * 86400
```
ahoym: **[MEDIUM] [python-engineer] `cutoff` is relative to the newest log entry, not wall-clock time**

`cutoff = max(s.get('ts', 0) for s in suggests) - args.days * 86400` anchors the lookback window to the most-recent suggestion in the log. If the log hasn't been updated in several days (e.g., the hook was temporarily disabled), `--days 7` won't cover the last 7 calendar days from now — it will cover 7 days before the last logged entry.

This may be intentional (relative to last activity), but it's a non-obvious semantic that will surprise future maintainers. Either:
- Add a comment: `# lookback from most-recent log entry, not wall clock`
- Or use `import time; cutoff = time.time() - args.days * 86400` for wall-clock semantics

---
- *Persona:* python-engineer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/analyze.py
```diff
@@ -0,0 +1,161 @@
...
+        p = home + p[1:]
+    p = str(Path(p).resolve()) if Path(p).exists() else p
+    # Strip the homedir prefix for readability
+    if p.startswith(home + "/"):
+        p = "~" + p[len(home):]
+    return p
+
+
+def main():
+    import argparse
```
ahoym: **[LOW] [python-engineer] `import argparse` deferred inside `main()` — non-idiomatic for stdlib**

`argparse` is stdlib, always available, and has no import side effects. Deferring it inside `main()` provides no benefit and makes top-level imports harder to audit. Move to the top-level import block.

---
- *Persona:* python-engineer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/analyze.py
```diff
@@ -0,0 +1,161 @@
...
+            if p not in suggested_paths_seen:
+                file_loaded_unsolicited[p] += 1
+
+    def panel(title):
+        print(f"\n{title}")
+        print("=" * len(title))
+
+    panel(f"Suggestion telemetry — last {args.days} days")
+    print(f"Total prompts logged:        {total_prompts}")
+    pct = (100 * prompts_with_suggestions // total_prompts) if total_prompts else 0
```
ahoym: **[LOW] [python-engineer] Integer division `//` truncates percentage display**

`(100 * prompts_with_suggestions // total_prompts)` truncates: `(100 * 1 // 3) == 33`, not `33%`. Same issue on the tier hit-rate line. For a human-readable analytics report, `round(100 * n / d)` is more accurate.

```python
pct = round(100 * prompts_with_suggestions / total_prompts) if total_prompts else 0
```

---
- *Persona:* python-engineer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Agreed. Extracted `Provider`, `home`, `expand_tilde`, and `providers(include_project_local: bool)` into `src/lib.rs`. `index-build.rs` now calls `providers(false)` with a comment documenting why projectLocal is intentionally omitted from the offline index (CWD-dependent, hook-runtime only). The behavioral split is now explicit at the call site rather than hidden in a duplicated function body. Will reply with the commit hash after push.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Agreed. `SCHEMA_VERSION` is now a single `pub const SCHEMA_VERSION: u64 = 1;` in `src/lib.rs`; both binaries import it via `learnings_suggest::SCHEMA_VERSION`. The type mismatch is gone by construction. Will reply with commit hash after push.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Agreed. Both moved to `src/lib.rs` as `pub fn home()` and `pub fn expand_tilde()`. `read-log.rs` also picks up the shared `home()` (it had its own copy too). Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Agreed. `anchor_commit(provs: &[Provider])` now takes the already-loaded slice; `main()` passes `&provs`. Saves a redundant file read + parse. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/main.rs
ahoym: Agreed. Extracted `match_sections`, `match_files`, `merge_and_dedup`, `compute_forced`, and `format_block`. `run()` is now a ~35-line coordinator: parse payload → run two matchers → merge + dedup → apply forced/score filter → format. Each phase has a single typed input/output (`SectionHits` and `FileHits` are now named type aliases) so they're individually testable. Smoke-tested with two prompts; output format is unchanged. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/main.rs
ahoym: Agreed — went with the comment option. Added a note above `extract_quoted` documenting that the matcher is ASCII-only and that macOS smart-quote autocorrect (U+201C/U+201D) silently bypasses the bypass. Broader Unicode-quote handling can be a follow-up if this surfaces in telemetry. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/bootstrap.sh
ahoym: Agreed on both. Added `[ "$bin" -nt "$DIR/Cargo.lock" ] || return 0` alongside the existing `Cargo.toml` check, and replaced the one-line comment with a fuller explanation of the `0=stale, return-0-short-circuits` inversion. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

setup-claude.sh
ahoym: Agreed. `setup-claude.sh` now delegates to `rebuild-index.sh` after the hook bootstrap, eliminating the third platform-case statement. The wrapper handles the silent no-op when no platform binary exists. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/settings.json
ahoym: Took option 2 (early-exit guard). Added `[ -f "$HOME/.claude/learnings-providers.json" ] || exit 0` to both `learnings-suggest.sh` and `read-log.sh`. This limits the blast radius to machines that have run `setup-claude.sh` — on non-dotfiles machines the wrapper short-circuits before any binary lookup.

Deferring option 1 (move registrations to `settings.local.json`) for now: that file is gitignored, so it would require new injection logic in `setup-claude.sh` and would change the contract for other dotfile consumers who want the hook globally. Happy to revisit if the guard proves insufficient in practice.

Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/guidelines/context-aware-learnings.md
ahoym: Agreed. Both rows now read `agent falls back to manual pipeline search when binary is absent`. The fallback path makes the keyword + domain-shift gates consistent with the pre-edit row's `Today: agent-owned` framing. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/guidelines/context-aware-learnings.md
ahoym: Took the tier-table conversion (collapsed 4 bullets into a 4-row table) and tightened the prose on the path-forms section. Kept the `offset=start, limit=(end-start+1)` formula — it's the specific conversion from the hint `path:start-end` form to Read parameters and I don't think it's redundant with CLAUDE.md's general "prefer offset+limit" guidance (which doesn't tell you how to translate a hint path).

Net reduction: ~10 lines (from ~42 to ~32 for the hook-suggested section). Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/analyze.py
ahoym: Agreed. Switched to wall-clock semantics: `cutoff = time.time() - args.days * 86400`. `time` is now imported at the top-level alongside `argparse`. The relative-to-newest-entry behavior was unintentional — wall-clock is what `--days 7` should mean. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/analyze.py
ahoym: Agreed. Moved `import argparse` to the top-level import block. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/analyze.py
ahoym: Agreed. Replaced `//` with `round(... / ...)` on all four percentage calls (overall pct, top-performers rate, plus both tier hit-rate lines). Re-running locally on a 2-of-3 case now prints `67%` instead of `66%`. Will reply with commit hash.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Fixed in da3836d.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Fixed in da3836d.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Fixed in da3836d.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Fixed in da3836d.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/main.rs
ahoym: Fixed in da3836d.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/main.rs
ahoym: Fixed in da3836d.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/bootstrap.sh
ahoym: Fixed in 28b6e79.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

setup-claude.sh
ahoym: Fixed in b9a0d0b.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/settings.json
ahoym: Fixed in 319f975.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/guidelines/context-aware-learnings.md
ahoym: Fixed in 17d8b79.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/guidelines/context-aware-learnings.md
ahoym: Fixed in 17d8b79.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/analyze.py
ahoym: Fixed in 035b294.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/analyze.py
ahoym: Fixed in 035b294.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/analyze.py
ahoym: Fixed in 035b294.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
```diff
@@ -0,0 +1,389 @@
...
+    lines: (usize, usize),
+    level: u8,
+    keywords: Vec<String>,
+}
+
+fn extract_explicit_keywords(file_text: &str) -> Vec<String> {
+    // Existing convention: line 2 of each file is `- **Keywords:** kw1, kw2, ...`
+    file_text
+        .lines()
+        .take(5)
```
ahoym: does this correspond with the standard header size? if so can we save this to a constant for readbability?

claude/hooks/learnings-suggest/src/bin/index-build.rs
```diff
@@ -0,0 +1,389 @@
...
+        .find_map(|line| {
+            let l = line.trim_start_matches(|c: char| c == '-' || c.is_whitespace());
+            l.strip_prefix("**Keywords:**")
```
ahoym: does this correspond with the standard header title? if so can we save this to a constant for readbability?

claude/hooks/learnings-suggest/src/bin/read-log.rs
```diff
@@ -0,0 +1,57 @@
...
+    let v: Value = serde_json::from_str(&s).ok()?;
+
+    if v.get("tool_name").and_then(Value::as_str)? != "Read" {
+        return None;
+    }
+    let input = v.get("tool_input")?;
+    let path = input.get("file_path").and_then(Value::as_str)?;
+
+    // Only log Reads of files that could plausibly be a learnings suggestion.
+    let relevant = ["/learnings/", "/guidelines/", "/skill-references/", "/commands/"]
```
ahoym: can we save this as a constant?

claude/hooks/learnings-suggest/src/bin/read-log.rs
```diff
@@ -0,0 +1,57 @@
...
+        .iter()
+        .any(|m| path.contains(m));
+    if !relevant {
+        return None;
+    }
+
+    let dir = home().join(".claude").join("claude-artifacts").join("ast");
```
ahoym: also save this as a constant with descriptive comments?

claude/hooks/learnings-staleness.py
```diff
@@ -0,0 +1,70 @@
...
+    if not (repo / ".git").exists():
+        return
+
+    try:
+        out = subprocess.check_output(
+            [
+                "git", "-C", str(repo), "diff", "--name-only",
+                f"{anchor}..HEAD", "--",
+                "claude/learnings/", "claude/guidelines/",
+                "claude/commands/", "claude/skill-references/",
```
ahoym: is it possible to save the list of specific directories to a constant?

the idea across the PR is making the scripts as "decoupled" to the internals of the scripts as much as possible

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Agreed — `5` is a search budget, not the header size. Extracting to a named constant (`KEYWORDS_HEADER_SEARCH_LIMIT`) with a comment explaining the bound. Will reply with commit ref.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Agreed — `**Keywords:**` is the repo-wide convention (line 2 of every learnings file). Pulling both forms into named constants (`KEYWORDS_MARKER` + `KEYWORDS_MARKER_LEGACY`) makes the convention explicit at the call site. Will reply with commit ref.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/read-log.rs
ahoym: Agreed. Extracting to a module-level `LEARNINGS_SUBPATHS` constant, mirroring the same set in `learnings-staleness.py` (different form, same concept — path fragments vs git-diff path args). Will reply with commit ref.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/read-log.rs
ahoym: Agreed. The same `~/.claude/claude-artifacts/ast` path is assembled in three Rust call sites (`main.rs`, `index-build.rs`, `read-log.rs`) plus `analyze.py`. Adding `pub fn artifacts_dir()` to `lib.rs` and routing all three Rust callers through it — same `decoupling` theme as comment #5. Will reply with commit ref.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-staleness.py
ahoym: Agreed — the "decoupling" framing lands. Extracting to a module-level `LEARNINGS_DIRS` tuple at the top of `learnings-staleness.py`, with a comment pointing at the mirror constant in `read-log.rs` so the two stay in mind together when the set changes. Will reply with commit ref.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Fixed in c455c46.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/index-build.rs
ahoym: Fixed in c455c46.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/read-log.rs
ahoym: Fixed in c455c46.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-suggest/src/bin/read-log.rs
ahoym: Fixed in c455c46.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

claude/hooks/learnings-staleness.py
ahoym: Fixed in c455c46.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7)*
- *Persona:* none
- *Role:* Addresser

ahoym (commented):
## Team Re-review: Add learnings-suggest hook: federated keyword matching & section-level indexing

`c455c46` extracts five inline magic literals across Rust and Python to named constants — `artifacts_dir()` in `lib.rs`, three keyword-scanning constants in `index-build.rs`, `LEARNINGS_SUBPATHS` in `read-log.rs`, and `LEARNINGS_DIRS` in `learnings-staleness.py`. Pure refactor with no logic changes; cross-reference comments between the mirrored Rust/Python constants are accurate and symmetric.

Reviewers (this cycle): orchestrator inline review (subagent-skip: <100 lines, pure constant extraction)
Carried forward: architecture-reviewer, claude-config-reviewer, python-engineer

### Previous Findings

- ✅ 14 resolved (verified in prior cycle — no change in status)

### New Findings

0 new findings.

### Positive Signals

- `artifacts_dir()` correctly deduplicates the three-part join chain (`home().join(".claude").join("claude-artifacts").join("ast")`) that previously appeared verbatim in `main.rs`, `index-build.rs`, and `read-log.rs`. The docstring cross-reference to `analyze.py`’s `ART` constant is accurate.
- The `LEARNINGS_SUBPATHS` ↔ `LEARNINGS_DIRS` cross-reference comments are symmetric and explain the form difference (leading `/` vs `claude/` prefix) — exactly what a future maintainer needs to keep them in sync.
- `KEYWORDS_HEADER_SEARCH_LIMIT` comment clarifies it’s a search budget, not a header-size assertion — the distinction matters if a file’s header is malformed.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (claude-sonnet-4-6)*
- *Persona:* architecture-reviewer, claude-config-reviewer, python-engineer
- *Role:* Team-Reviewer

claude/hooks/learnings-suggest/bootstrap.sh
```diff
@@ -0,0 +1,114 @@
...
+    [ -z "$(find "$DIR/src" -type f -newer "$bin" -print -quit 2>/dev/null)" ] || return 0
+  done
+  return 1
+}
+
+ensure_cargo() {
+  command -v cargo >/dev/null 2>&1 && return 0
+  echo "Installing Rust toolchain..."
+  if [ "$OS" = "Darwin" ] && command -v brew >/dev/null 2>&1; then
+    brew install rust
```
ahoym: can we actually use `rustup` for mac os too? it's best practice to use a version manager

ahoym: Agreed — switching to `rustup` on macOS too. Two reasons:

1. **Best-practice alignment.** `rustup` is the official Rust toolchain manager from rust-lang.org and the recommended install path across all platforms, including macOS. It gives us stable/beta/nightly switching, `rust-toolchain.toml` support, and a consistent install surface across machines.
2. **Latent bug in `--all-targets`.** The cross-build path at line 103 calls `rustup target add "$triple"`. If a macOS user installs Rust via `brew install rust`, `rustup` isn't on PATH and `--all-targets` silently no-ops the target install (the `|| true` swallows it), then `cargo zigbuild --target …` fails on the missing toolchain. Unifying on `rustup` removes the divergence.

Will collapse `ensure_cargo()` to the single `rustup` branch (drop the `brew install rust` arm) and reply with the commit hash after push.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

ahoym: Fixed in 7e5c174.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

ahoym: Course-correction — kept `rustup` as the toolchain but installed it via `brew install rustup` on macOS so updates are tracked by the package manager. The keg-only `rustup` formula ships `rustup-init` at `$(brew --prefix rustup)/bin/rustup-init`; running it seeds `~/.cargo/bin` exactly like the curl path, so the rest of the script (and `--all-targets`) is unchanged. Non-macOS / non-brew machines still fall through to the upstream `sh.rustup.rs` installer.

Fixed in ebe414d.

---
- *Co-Authored with [Claude Code](https://claude.ai/code) (Claude Opus 4.7 (1M context))*
- *Persona:* none
- *Role:* Addresser

